PR #5, opened 2024-11-14
Simple binary encoding and decoding

govereau:
A prototype for a binary serialization format.
Proof of concept for a specification document.


NKL/Encode.lean
```diff
@@ -0,0 +1,331 @@
+/-
```
seanmcl, line 1: Would it make sense to have a more user friendly version of this as either a either a big comment or separate doc?

NKL/Encode.lean
```diff
@@ -0,0 +1,331 @@
...
+
+private def decInt' : DecodeM (Bool × Nat) := do
+  let sign <- next
+  let n <- decNat
+  return (sign == 1, n + sign.toNat)
+
+local instance : BEq ByteArray where
+  beq a b := a.data == b.data
+
+#guard encInt 1    == .mk #[0, 1, 0, 0, 0, 0, 0, 0, 0]
```
seanmcl, line 108: Pedantic comment: imo it's a bit unusual to have the sign bit as the 0th bit on a little-endian representation. The sign seems like the "most important" bit.

NKL/Encode.lean
```diff
@@ -0,0 +1,331 @@
...
+-- TODO:
+-- I am keeping the encode functions pure; encFloat can fail,
+-- but this seems unlikely as the Lean parser would have to not understand
+-- the result of toString on a Float value
+
+private def encFloat (f : Float) : ByteArray :=
+  let str := (toString f).toSubstring
+  let one : String.Pos := { byteIdx := 1 }
+  let sgn := str.get 0 == '-'
+  let str := bif sgn then { str with startPos := one } else str
```
seanmcl, line 134: Why `bif` instead of `if`?

NKL/Encode.lean
```diff
@@ -0,0 +1,331 @@
...
+
+#guard decode' decInt' (encInt 1) == some (false, 1)
+#guard decode' decInt' (encInt 0) == some (false, 0)
+#guard decode' decInt' (encInt (-0)) == some (false, 0)
+#guard decode' decInt' (encInt $ -1) == some (true, 1)
+
+------------------------------------------------------------------------------
+-- Float values are encoded as (mantissa, exponent), where the
+-- mantissa and exponenet are integers.
+-- TODO:
```
seanmcl, line 125: Can we have a TODO to switch to using bits when Lean supports it? Would like to not get any of these stored before that happens!

NKL/Encode.lean
```diff
@@ -0,0 +1,331 @@
...
+#guard decode' decFloat (encFloat 0.5) == some 0.5
+#guard decode' decFloat (encFloat 1.0) == some 1.0
+#guard decode' decFloat (encFloat 1.5) == some 1.5
+#guard decode' decFloat (encFloat (-0)) == some (0.0)
+#guard decode' decFloat (encFloat (-0.5)) == some (-0.5)
+#guard decode' decFloat (encFloat (-1.0)) == some (-1.0)
+#guard decode' decFloat (encFloat (-1.5)) == some (-1.5)
+
+------------------------------------------------------------------------------
+-- Strings are encoded as a length followed by a sequence of UTF8 bytes
```
seanmcl, line 164: Just reading the comment, I wonder how do you know how many bytes the length is?  I think this is bc you are using Nat in place of UInt64?

NKL/Encode.lean
```diff
@@ -0,0 +1,331 @@
...
+   , UInt8.ofNat $ n >>> 8
+   , UInt8.ofNat $ n >>> 16
+   , UInt8.ofNat $ n >>> 24
+   , UInt8.ofNat $ n >>> 32
+   , UInt8.ofNat $ n >>> 40
+   , UInt8.ofNat $ n >>> 48
+   , UInt8.ofNat $ n >>> 56
+   ]
+
+private def decNat : DecodeM Nat := do
```
seanmcl, line 63: Could we do `UInt64` here so there is no question about the number of bytes?

seanmcl (commented):
Nice!

NKL.lean
```diff
@@ -5,3 +5,4 @@ Authors: Paul Govereau
 -/
 import NKL.NKI
 import NKL.FFI
+import NKL.Encode
```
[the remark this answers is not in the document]
govereau: Yeah, we probably should.

NKL/Encode.lean
govereau: I am not sure what you mean. A more friendly version of what?

NKL/Encode.lean
govereau: Yeah, I think I will switch the AST to use Int64 (upgrading to lean 4.14.0)

NKL/Encode.lean
govereau: Indeed, also 9 bytes is a too many bytes :)
Having Int64 in lean will fix this.

NKL/Encode.lean
govereau: Yes! Agreed.

NKL/Encode.lean
govereau: I was just following something I read in the lean manual. `bif` is specialized to boolean conditions; not sure if there is really any difference.

NKL/Encode.lean
govereau: I don't really, if you had a string longer that 2^64, this code will not work correctly. But, then I guess `encString` needs to be able to fail... Perhaps both encode and decode should be in a monad?

NKL/Encode.lean
seanmcl: Oh, just the formatter. I was imagining a bnf-looking sort of thing without all the monads. But maybe it's fine to not have it in a doc for now.

seanmcl (approved):

PR #7, opened 2024-11-20
Sketch of Operational Semantics for Expressions

govereau:


NKL/Oper.lean
```diff
@@ -0,0 +1,181 @@
...
+-- Some experimental definitions of semantics for expresssions
+-- For now, I am abstracting over the types of tensors and
+-- tensor operators.
+
+-- Evaluating expressions results in values, which
+-- may contain tensors and tensor operators
+inductive Value (tensor op : Type) where
+  | const (c : Const)
+  | tensor (t : tensor) (ix : List (Value tensor op))
+  | slice (l u step : Value tensor op)
```
seanmcl, line 20: Chatted offline. `step` could be an int here, but in NumPy is also allowed to be a singleton "scalar" array like `np.array(1)`. Not sure which type is better here.

NKL/Oper.lean
```diff
@@ -0,0 +1,181 @@
...
+  | oper (f : op)
+  | tuple (l : List (Value tensor op))
+  | list (l : List (Value tensor op))
```
seanmcl, line 23: Do we need both tuples and lists?

NKL/Oper.lean
```diff
@@ -0,0 +1,181 @@
...
+-- gridcall not included: this likely needs to be a statement
+
+mutual
+inductive StepExpr [Env E] : (env : E) -> Expr -> Val E -> Prop
+ | bvar :
+     Env.lookup? env a = some v ->
+     StepExpr env (.bvar a) v
+ | var :
+     Env.lookup env a = v ->
+     StepExpr env (.var a b) v
```
seanmcl, line 54: What happens to `b` here? Is `b` the initial value that is then just carried around and ignored?

NKL/Oper.lean
```diff
@@ -0,0 +1,181 @@
...
+ | subscript :
+     StepExpr env e (.tensor t []) ->
```
seanmcl, line 56: Why `[]`?

NKL/Oper.lean
```diff
@@ -0,0 +1,181 @@
...
+     StepList env le lv ->
+     StepExpr env (.subscript e le) (.tensor t lv)
+ | slice :
+     StepExpr env l l' ->
+     StepExpr env u u' ->
+     StepExpr env step step' ->
+     StepExpr env (.slice l u step) (.slice l' u' step')
+ | binop :
```
seanmcl, line 64: Why do we have both `binop` and `call`?

seanmcl (commented):
Nice. Can see it coming together.

NKL/Oper.lean
```diff
@@ -0,0 +1,181 @@
...
+
+namespace NKL
+
+-- Some experimental definitions of semantics for expresssions
+-- For now, I am abstracting over the types of tensors and
+-- tensor operators.
+
+-- Evaluating expressions results in values, which
+-- may contain tensors and tensor operators
+inductive Value (tensor op : Type) where
```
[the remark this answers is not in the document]
govereau: `op` is the type of operator values. So, if I look up "matmul" in the environment, it will return something that I can give to apply. I don't need to know what the something is, so it is abstract.

NKL/Oper.lean
govereau: I guess, in the current syntax, this might be written like so: `u=np.array(1); t[1:2:u]`. It is weird to have just `u` and not `u[0]`. Will need to consider how to handle this, well.

NKL/Oper.lean
govereau: No, probably not, it is just a left-over from the python syntax.

NKL/Oper.lean
govereau: Also, left-over from the python syntax. Perhaps we should define a cleaner core language and translate the initial python syntax down to remove this kind of stuff?

NKL/Oper.lean
govereau: Great eye. I was going to bring this up. I have never seen expressions like `t[a,b,...][x,y,...]` in NKI kernels, and I am not sure it is allowed. That is, you have to write the indexes all at once `t[a,b,...,x,y,..]`, and you can't subscript into a partial tensor. Of course, allowing this kind of nested indexing is more general, and maybe we should?

NKL/Oper.lean
govereau: Another left over from python syntax. We could translate the `binop` to `call` and the `gridcall` to `forloop` and `call`... and `cond` can be eliminated, as can `var` and `check`. More arguments for a "core language".

seanmcl (approved):

PR #8, opened 2024-12-09
feat: support for python syntax

govereau:
Introduces a direct representation of the python abstract syntax, and a simple parser for python functions. This is necessary to support the formalization of the tracing process.

### DIFF
--- a/NKL/FFI.lean
+++ b/NKL/FFI.lean
@@ -6,16 +6,24 @@ Authors: Paul Govereau
 import Lean
 import NKL.NKI
 import NKL.PrettyPrint
+import NKL.Python
 
 namespace NKL
 
--- temporary for testing
+local instance : MonadLift (Except String) IO where
+  monadLift
+    | .ok x => return x
+    | .error s => throw $ .userError s
+
+@[export parse_json_old]
+def parse_json_old (json : String) : IO Unit := do
+  let jsn <- Lean.Json.parse json
+  let f:Fun <- Lean.fromJson? jsn
+  print_nki f
 
 @[export parse_json]
-def parse_json (json : String) : IO Unit := do
-  match Lean.Json.parse json with
-  | .error str => throw $ .userError str
-  | .ok jsn => do
-    match Lean.fromJson? jsn with
-    | .error str => throw $ .userError str
-    | .ok (f:Fun) => print_nki f
+def parse_json (s : String) : IO Unit := do
+  let kernel <- Python.Parsing.parse s
+  for (n,f) in kernel.funcs do
+    IO.println s!"found {n}"
+    IO.println s!"{repr f}"
--- a/NKL/Python.lean
+++ b/NKL/Python.lean
@@ -1,0 +1,288 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Paul Govereau
+-/
+import Lean
+
+/-!
+# Abstract syntax of Python functions
+
+Mostly 1-to-1 translation of the Python AST to lean.
+see: https://docs.python.org/3/library/ast.html
+-/
+
+namespace NKL
+namespace Python
+
+deriving instance Repr for Lean.JsonNumber
+
+structure Pos where
+  lineno : Nat
+  end_lineno : Nat := 0
+  col_offset : Nat := 0
+  end_col_offset : Nat := 0
+  deriving Repr
+
+inductive Const where
+  | none
+  | bool (value: Bool)
+  | num (value: Lean.JsonNumber)
+  | string (value: String)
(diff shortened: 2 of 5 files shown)


NKL/FFI.lean
```diff
@@ -6,16 +6,24 @@ Authors: Paul Govereau
 import Lean
 import NKL.NKI
 import NKL.PrettyPrint
+import NKL.Python
 
 namespace NKL
 
--- temporary for testing
+local instance : MonadLift (Except String) IO where
```
seanmcl, line 13: `local instance`! Nice.

NKL/Python.lean
```diff
@@ -0,0 +1,288 @@
...
+# Abstract syntax of Python functions
+
+Mostly 1-to-1 translation of the Python AST to lean.
+see: https://docs.python.org/3/library/ast.html
+-/
+
+namespace NKL
+namespace Python
+
+deriving instance Repr for Lean.JsonNumber
```
seanmcl, line 18: Wow, also didn't know you can derive instances far from the declaration.

NKL/Python.lean
```diff
@@ -0,0 +1,288 @@
...
+inductive Const where
+  | none
+  | bool (value: Bool)
+  | num (value: Lean.JsonNumber)
+  | string (value: String)
+  | ellipsis
+  deriving Repr
+
+inductive Ctx where
+  | load | store | del
```
seanmcl, line 36: Do we need these? I'd imagine we could/should do our own def/use analysis?

NKL/Python.lean
```diff
@@ -0,0 +1,288 @@
...
+  vararg : Option String
+  kwonlyargs : List String
+  kw_defaults: List Expr
+  kwarg : Option String
+  deriving Repr
+
+structure Fun where
+  source : String
+  args : Args
+  defaults: List Const
```
seanmcl, line 96: Is it obvious which of the `Args` these defaults apply to?

NKL/Python.lean
```diff
@@ -0,0 +1,288 @@
...
+    | .error s => throw s
+
+private def str : Json -> Parser String :=
+  monadLift ∘ Json.getStr?
+
+private def field (f: Json -> Parser a) (j : Json) (name : String) : Parser a :=
+  j.getObjVal? name >>= f
+
+private def field? (f: Json -> Parser a) (j : Json) (name : String) : Parser (Option a) :=
+  try let x <- field f j name; return (some x)
```
seanmcl, line 129: Is this the same as
```
try field f j name catch _ => return none
```
? If that's right, I'm fine with this, but it seems like a lot of the code is going for maximally point free style.

NKL/Python.lean
```diff
@@ -0,0 +1,288 @@
...
+    | _ => throw s!"unsupported python construct {key}"
+
+  keyword (j: Json) : Parser Keyword := do
+    let j <- j.getObjVal? "keyword"
+    return ⟨ <- field str j "arg", <- field expr j "value", <- pos j ⟩
+
+partial def stmt (j : Json) : Parser Stmt :=
+  withPos stmt' Stmt.stmtPos j
+where
+  stmt' (key : String) (j : Json) : Parser Stmt' := do
```
seanmcl, line 233: When should we prefer 

```
def stmt ...
  let stmt' ... = ...
```

vs

```
def stmt ...
  E
where
  ...
```
?

interop/nkl/parser.py
```diff
@@ -0,0 +1,134 @@
...
+import json
+
+from textwrap import dedent
+from itertools import chain
+from collections import deque
+
+class Enc(json.JSONEncoder):
+  def default(self, obj):
+    if isinstance(obj, ast.AST):
+      if len(obj.__dict__) == 0:
```
seanmcl, line 17: This just seems like magic. Should we add comments here saying what is going on, or is it obvious? Comment applies to the whole file. Maybe an example as a comment would help the reader.

NKL/Python.lean
govereau: We don't, but I am keeping all the python stuff intact to avoid having to modify it on the python side.

NKL/Python.lean
govereau: No, it isn't :)
I will put a comment, but you basically have to read the python AST documentation.

NKL/Python.lean
govereau: I think the `return (some x)` is needed to get the types right. Maybe `field f j name >>= return . some` :)

NKL/Python.lean
govereau: Good question. I am not sure. I see both styles used in the Lean compiler without any clear rhyme or reason.

interop/nkl/parser.py
govereau: I will add some comments. It is fairly opaque.

PR #9, opened 2024-12-11
chore: update python tests to use new parser

govereau:
Cleanup python tests; add some comments.


interop/nkl/parser.py
```diff
@@ -10,6 +10,25 @@
...
+# dictionary is empty we just return the class name.
+# e.g.
+# Binop(left=l,op=o,right=r), becomes:
+#    { BinOp : { left:l, op:o, right:r } }
+# Pass(), becomes
+#    'Pass'
+#
+# For anything else not handled by the default
+# encoder, we return "...", the Ellipsis.
+# Conveniently, Ellipsis is one of the things
```
seanmcl, line 28: Won't we need ellipses for indexing?

seanmcl (approved):
Thank you for the comments.

PR #10, opened 2024-12-28
feat: add global references and arguments

govereau:
Add ability to capture global references, including function arguments when parsing kernels. Globals and arguments are directly encoded as KLR terms.

### DIFF
--- a/NKL/Python.lean
+++ b/NKL/Python.lean
@@ -15,10 +15,8 @@ see: https://docs.python.org/3/library/ast.html
 namespace NKL
 namespace Python
 
-deriving instance Repr for Lean.JsonNumber
-
 structure Pos where
-  lineno : Nat
+  lineno : Nat := 0
   end_lineno : Nat := 0
   col_offset : Nat := 0
   end_col_offset : Nat := 0
@@ -26,14 +24,15 @@ structure Pos where
 
 inductive Const where
   | none
-  | bool (value: Bool)
-  | num (value: Lean.JsonNumber)
-  | string (value: String)
+  | bool (value : Bool)
+  | int (value : Int)
+  | float (value : Float)
+  | string (value : String)
   | ellipsis
   deriving Repr
 
--- We don't need these, but we preserve them to make copying
--- the Python AST over easier.
+-- This context comes from the Python AST.
+-- The store hint is used by the tracing implementation for simplicity.
 inductive Ctx where
   | load | store | del
   deriving Repr
@@ -44,8 +43,9 @@ inductive Expr where
   deriving Repr
 
 inductive Expr' where
-  | const (value: Const)
-  | name (id: String) (ctx : Ctx)
+  | const (value : Const)
+  | tensor (shape : List Expr) (dtype : String)
+  | name (id : String) (ctx : Ctx)
   | attr (value : Expr) (id : String) (ctx : Ctx)
   | tuple (xs: List Expr) (ctx : Ctx)
   | list (xs: List Expr) (ctx : Ctx)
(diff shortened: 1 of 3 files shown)


NKL/Python.lean
```diff
@@ -44,8 +43,9 @@ inductive Expr where
   deriving Repr
 
 inductive Expr' where
-  | const (value: Const)
-  | name (id: String) (ctx : Ctx)
+  | const (value : Const)
+  | tensor (shape : List Expr) (dtype : String)
```
seanmcl, line 47: Reserving judgment until reading more, but not obvious to me how this will be used. Maybe some documentation would be good.

NKL/Python.lean
```diff
@@ -124,10 +131,22 @@ structure Fun where
...
+    field `kwargs`
+  - global variables referenced by any of the functions
+-/
 structure Kernel where
   entry : String
   funcs : List (String × Fun)
-  globals : List (String × Option String)
+  args : List Expr'
+  kwargs : List (String × Expr')
+  globals : List (String × Expr')
```
seanmcl, line 149: Nice documentation string. What's an example of a global? Where does it come from?

NKL/Python.lean
```diff
@@ -191,29 +210,30 @@ private def withPos (p : String -> Json -> Parser b) (f : b -> Pos -> a) : Json
     return (f exp pos)
   | _ => throw "expecting object"
 
+def genError (source err : String) (pos : Pos) : String :=
```
seanmcl, line 213: Maybe an example of what you're parsing here would be nice.

NKL/Python.lean
```diff
@@ -302,14 +322,40 @@ def function (j : Json) : Parser Fun := do
     let body <- field (list stmt) j "body"
     return Fun.mk source args defaults body
 
+-- Both global references and arguments are processed in the global
+-- environment. These terms do not have a position, and must be
+-- evaluable in the default environment.
+partial def global : Json -> Parser Expr'
+  | .null => return .const .none
+  | .obj (.node _ _ "fun"   (.str  s)    _) => return .name s .load
```
seanmcl, line 330: I still don't understand what `.load` does.

interop/nkl/parser.py
```diff
@@ -42,19 +43,52 @@ def default(self, obj):
     except Exception:
       return "..."
 
+# Referenced names, that are not functions are placed in the
+# global environment. Unlike functions, these values cannot
+# reflected on using the ast module (the inspect module can only
```
seanmcl, line 48: grammar

seanmcl (approved):

interop/nkl/parser.py
govereau: thanks!

NKL/Python.lean
```diff
@@ -15,25 +15,24 @@ see: https://docs.python.org/3/library/ast.html
...
+  | int (value : Int)
+  | float (value : Float)
+  | string (value : String)
   | ellipsis
   deriving Repr
 
--- We don't need these, but we preserve them to make copying
--- the Python AST over easier.
+-- This context comes from the Python AST.
+-- The store hint is used by the tracing implementation for simplicity.
```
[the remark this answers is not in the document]
govereau: I will add some comment in a new PR.

NKL/Python.lean
govereau: Yes, this may need to change as more kernels are supported in the tracer. So far, I am able to get by with this basic type information.

NKL/Python.lean
govereau: Will add a comment.

PR #11, opened 2024-12-30
feat: define KLR representation

govereau:
This is the portable format, and the target of tracing.

### DIFF
--- a/NKL/KLR.lean
+++ b/NKL/KLR.lean
@@ -1,0 +1,107 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Paul Govereau
+-/
+
+
+/-!
+# Abstract syntax of Core NKL language
+
+This language is the result of "tracing", and is used as the
+portable format, a.k.a. Kernel Language Representation (KLR).
+-/
+
+namespace NKL.KLR
+
+-- TODO
+inductive Ty where
+
+inductive Const where
+  | none
+  | bool (value : Bool)
+  | int (value : Int)
+  | float (value : Float)
+  | string (value : String)
+  deriving Repr, BEq
+
+namespace Const
+
+-- Python-like rules for conversion to boolean
+def isTrue : Const -> Bool
+  | .none     => false
+  | .bool b   => b
+  | .int i    => i != 0
+  | .float f  => f != 0.0
+  | .string s => s != ""
+
+-- Python-like rules for conversion to integer
+def toInt : Const -> Except String Int
+  | .none       => throw "none cannot be converted to an integer"
+  | .bool true  => return 1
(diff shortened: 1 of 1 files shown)


NKL/KLR.lean
```diff
@@ -0,0 +1,107 @@
...
+namespace NKL.KLR
+
+-- TODO
+inductive Ty where
+
+inductive Const where
+  | none
+  | bool (value : Bool)
+  | int (value : Int)
+  | float (value : Float)
```
seanmcl, line 24: Will we eventually need Float32/Float16/Float8 etc?

NKL/KLR.lean
```diff
@@ -0,0 +1,107 @@
...
+def toInt : Const -> Except String Int
+  | .none       => throw "none cannot be converted to an integer"
+  | .bool true  => return 1
+  | .bool false => return 0
+  | .int i      => return i
+  | .float f    =>
+      -- Python is a bit strange here, it truncates both
+      -- positive and negative numbers toward zero
+      if f < 0.0 then
+        return (Int.ofNat (Float.floor (-f)).toUInt64.toBitVec.toNat).neg
```
seanmcl, line 48: `toUInt64.toBitVec.toNat` could go in some common library.

NKL/KLR.lean
```diff
@@ -0,0 +1,107 @@
...
+      else
+        return Int.ofNat (Float.floor f).toUInt64.toBitVec.toNat
+  | .string s   =>
+      match s.toInt? with
```
seanmcl, line 52: Seems unlikely python and lean match here?

NKL/KLR.lean
```diff
@@ -0,0 +1,107 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Paul Govereau
+-/
+
+
+/-!
+# Abstract syntax of Core NKL language
+
+This language is the result of "tracing", and is used as the
+portable format, a.k.a. Kernel Language Representation (KLR).
```
seanmcl, line 12: Some examples would be nice. Maybe another PR.

seanmcl (approved):

NKL/KLR.lean
govereau: I am not sure. These expressions are, conceptually, python expressions. So, really this should be a big float?
It is only when these constants reach a tensor that we need to check if they are compatible.

NKL/KLR.lean
govereau: They seem to match based on a some playing around in the interpreter. However, yeah, there are probably some corner cases. This will be an issue if someone writes a kernel that uses the implicit string to int conversion in python.

NKL/KLR.lean
govereau: Agreed.

NKL/KLR.lean
govereau: The tracing patches have some examples.

PR #13, opened 2024-12-30
chore: remove unused code

govereau:
Remove unused code, and added some comments as suggested in prior reviews.


seanmcl (approved):
Nice comments

PR #16, opened 2025-01-08
fix: handle arguments and defaults consistently

govereau:
Default function arguments were not being handled the same way as function arguments. It was possible that an expression that could be passed as an argument could not have been used as a default value for the same argument. This patch fixes this issue, and simplifies some of the parsing code for arguments and function signatures.


NKL/Python.lean
```diff
@@ -129,16 +129,14 @@ def Args.names (ax : Args) : List String :=
...
 
-/-
-In addition to the defaults above from the AST, we also collect
-the values from f.__defaults__ here in the Fun structure. These
-values are evaluated in a different context from the other names
-in the function, so we need to capture them on the Python side.
--/
+def Args.all_defaults (ax : Args) : List (String × Expr') :=
+  let args := ax.posonlyargs ++ ax.args
+  let dflt := args.reverse.zip ax.defaults.reverse
```
seanmcl, line 134: Nit: I'd find `(args.zip defaults).reverse` marginally easier to grok at a glance.

NKL/Python.lean
```diff
@@ -129,16 +129,14 @@ def Args.names (ax : Args) : List String :=
   let xs := match ax.kwarg  with | none => xs | some x => xs.append [x]
   xs
 
-/-
-In addition to the defaults above from the AST, we also collect
-the values from f.__defaults__ here in the Fun structure. These
-values are evaluated in a different context from the other names
-in the function, so we need to capture them on the Python side.
--/
+def Args.all_defaults (ax : Args) : List (String × Expr') :=
```
seanmcl, line 132: Just curious, what does "ax" stand for in your head?

NKL/Python.lean
```diff
@@ -129,16 +129,14 @@ def Args.names (ax : Args) : List String :=
...
+  let args := ax.posonlyargs ++ ax.args
+  let dflt := args.reverse.zip ax.defaults.reverse
+  dflt ++ ax.kw_defaults
```
seanmcl, line 135: Is it ever possible that the list has multiple copies of the same name? If so is it OK? If not, why not?

seanmcl:
Could you give an example of
> It was possible that an expression that could be passed as an argument could not have been used as a default value for the same argument.
?

seanmcl (approved):

NKL/Python.lean
govereau: Ah, the lists are not necessarily the same length. If we have args [a,b,c], and defaults [1,2] (this is from the python parser), then the meaning is that the default for b=1 and c=2. The argument a must be provided by caller.

NKL/Python.lean
govereau: Oh, should be `args`, not sure why I used `ax`. In my head this is a collection, not an arg list or arg dict, but something more complex "arg X".

NKL/Python.lean
govereau: It shouldn't be posssible. To get this you would need, e.g.
```
def f(x,x): ...
```
Which would be a syntax error in python.

NKL/Python.lean
seanmcl: Cool, a comment would be nice here. That was not at all obvious to me.

NKL/Python.lean
seanmcl: Ah, I kept forgetting that we were talking about Python semantics here.

PR #17, opened 2025-01-08
refactor: cleanup KLR definitions

govereau:
Remove some unnecessary parts of KLR, and reorganize the source files.


NKL/KLR/Basic.lean
```diff
@@ -0,0 +1,104 @@
+/-
```
seanmcl, line 1: I thought that the Lean convention is for the `Basic` file to just be a list of imports. How do you think of `Basic.lean` files?

NKL/KLR/Basic.lean
```diff
@@ -0,0 +1,104 @@
...
+
+abbrev Dtype := String
+abbrev Shape := List Int
+structure Tensor where
+  dtype : Dtype
+  shape : Shape
+  deriving Repr, BEq
+
+-- TODO
+inductive Typ where
```
seanmcl, line 32: What is this?

NKL/KLR/Basic.lean
```diff
@@ -0,0 +1,104 @@
...
+def toInt : Const -> Except String Int
+  | .none       => throw "none cannot be converted to an integer"
+  | .bool true  => return 1
+  | .bool false => return 0
+  | .int i      => return i
+  | .float f    =>
+      -- Python is a bit strange here, it truncates both
+      -- positive and negative numbers toward zero
+      if f < 0.0 then
+        return (Int.ofNat (Float.floor (-f)).toUInt64.toBitVec.toNat).neg
```
seanmcl, line 62: This is the same as `....toUInt64.toNat`

NKL/KLR/Basic.lean
```diff
@@ -0,0 +1,104 @@
...
+      else
+        return Int.ofNat (Float.floor f).toUInt64.toBitVec.toNat
+  | .string s   =>
+      match s.toInt? with
```
seanmcl, line 66: Hmm, I doubt python and Lean match up here? Maybe just add a comment that we're aware of the discrepancy?

NKL/KLR/Basic.lean
```diff
@@ -0,0 +1,104 @@
...
+
+end Const
+
+inductive IndexExpr where
+  | var (name : String)
+  | int (i : Int)
+  | neg (expr : IndexExpr)
+  | add (left right : IndexExpr)
+  | mul (scalar : Int) (expr : IndexExpr)
+  | floor (expr : IndexExpr) (scalar : Int)
```
seanmcl, line 78: What does `floor` do with the scalar? Is this obvious or should there be a comment?

NKL/KLR/Basic.lean
```diff
@@ -0,0 +1,104 @@
...
+  | ceil (expr : IndexExpr) (scalar : Int)
+  | mod (expr : IndexExpr) (scalar : Int)
+  deriving Repr, BEq
+
+-- Note: `np.newindex` is represented as `(.coord none)`
+inductive Index where
+  | ellipsis
+  | coord (e : Option IndexExpr)
```
seanmcl, line 86: Why option? Is `none` like `np.newaxis`?

NKL/KLR/Basic.lean
```diff
@@ -0,0 +1,104 @@
...
+  | slice (l u step : Option IndexExpr)
+  deriving Repr, BEq
+
+inductive Expr where
+  | var (x : String)
+  | const (c : Const)
+  | tensor (t : Tensor)
+  | access (t : Expr) (ix : List Index)
+  | call (f : Expr) (args : List Expr) (kwargs : List (String × Expr))
```
seanmcl, line 95: Do we want the `kwargs` at this level? Seems like they should be compiled away at some point pretty high up.

NKL/KLR/Basic.lean
```diff
@@ -0,0 +1,104 @@
...
+  deriving Repr, BEq
+
+inductive Stmt where
+  | pass
+  | expr (v : Expr)
```
seanmcl, line 100: Why is `expr` a statement in a language without side effects?

NKL/KLR/Basic.lean
seanmcl: Just looked around in other blessed repos and found Basic files with non-import content like data types.

NKL/KLR/Basic.lean
govereau: I think we could have two type systems. There is one defined over Terms used in tracing, but this is more python-like (it has any), and just needs to be good enough to hint what the tracing should do. Once we get to KLR, we could give a precise type system. It is not clear, to me, how useful this will be. This is just a reminder to myself that KLR is an untyped language right now.

NKL/KLR/Basic.lean
govereau: Oh, cool I will update the code.

NKL/KLR/Basic.lean
govereau: I wasn't able to find a counter-example. I will add some comments about this, I could have easily missed something here.

NKL/KLR/Basic.lean
govereau: I will add a comment; it is just normal integer division. Maybe it should be called `div`, but I was following Pellican.

NKL/KLR/Basic.lean
govereau: This is a peculiarity of the parser. If a user writes `t[None,:]`, this turns into `[coord none, slice none none none]`. I think this could also be written `t[np.newaxis,:]`, which would come out the same. The parser doesn't distinguish between the nones inside the slice and the ones outside the slice, so the new axis gets mixed in with the coord alternative.

NKL/KLR/Basic.lean
govereau: Totally agree. They should be removed by this point. I think it is not too difficult, I just need something like `bind_args` but over TermType. Also, we may need to specify types for all of the Global builtins to have enough information not to get stuck.

A related API design question is, if a NKI API has a signature like:
```
# from nki.isa
def iota(expr, dtype, mask=None, **kwargs): ...
```
What are we supposed to make of the kwargs here? I think the intention is that, in the future, maybe this ISA primitive could take some more arguments, and no changes to the signature are required. Or maybe there are undocumented flags you can pass to the primitive in kwargs. If so, then how do we represent this in KLR? In general, we may not be able to remove the kwargs from KLR if the API is handled this way.

Note, in the current NKL implementation, if we see, e.g.
```
nisa.iota(e, dty, foo="bar")
```
We fail with a message like "var args is not supported". This is because I am hoping we can eliminate these kwargs before we get to KLR, but we need some clarity on how the APIs will be evolved to make the call.

NKL/KLR/Basic.lean
govereau: Very good question. The (admittedly bad) answer is that I am not sure KLR is the final step. Maybe it isn't pure, but there is a transform we can do to make it pure. For example, we see stuff like:
```
# in nki.isa
 def builtin_custom_op(fn_name, file_name, ucode_version, isa_version, srcs, dsts, **kwargs):
```
If you call this function, it would look like:
```
def kernel(...):
  ...
  builtin_custom_op("name", "file", 1, 1, a, b)
  return b
```
Right, now the call to `builtin_custom_op` would be one of these expression statements. Not sure what to do here?

NKL/KLR/Basic.lean
govereau: I am not sure. I am happy to use any convention.

NKL/KLR/Basic.lean
seanmcl: oh intersting, so python adds the `coord`?

NKL/KLR/Basic.lean
seanmcl: I guess KLR could take all the arguments, and during the compilation we drop anything that doesn't map to a KLR argument name? Presumably at time T, KLR would handle any kwarg in the isa instruction. Then if a new arg is added to the ISA, we are somehow notified and catch up asap, in the meantime just quietly dropping the new argument in kernels.

govereau: OK, I think we agree. I will do a follow-up PR that removes the kwargs from KLR.

PR #18, opened 2025-01-08
Tracing type definitions

govereau:
feat: basic definitions to support tracing


NKL/Trace/Types.lean
```diff
@@ -0,0 +1,268 @@
...
+
+This module defines types to represent the built-ins, the environments, and the
+tracing monads.
+-/
+
+namespace NKL.Trace
+open NKL.KLR
+
+-- Lean already has a perfectly nice hierarchical string type
+export Lean (Name)
```
seanmcl, line 65: Ha, neat. Never seen `export` before.

NKL/Trace/Types.lean
```diff
@@ -0,0 +1,268 @@
...
+deriving instance Ord for Name
```
seanmcl, line 66: What is the scope of this `Ord` instance? E.g. do I need to `open NKL.Trace` for it to be in scope?

NKL/Trace/Types.lean
```diff
@@ -0,0 +1,268 @@
...
+
+-- Terms are an extension of KLR.Expr, and they have types, which may be `any`.
+-- TODO: can we get rid of any?
+
+inductive TermType where
+  | none | bool | int | float | string
+  | any    : Name -> TermType
+  | tuple  : List TermType -> TermType
+  | list   : List TermType -> TermType
+  | tensor : Dtype -> Shape -> TermType
```
seanmcl, line 78: Another choice is to have `tensor : Tensor -> TermType` here.

NKL/Trace/Types.lean
```diff
@@ -0,0 +1,268 @@
...
+  deriving Repr, BEq
+
+mutual
+structure Builtin where
+  name : Name
+  type : TermType
+  attr : String -> Except String Term
```
seanmcl, line 85: Could we use a simpler data type than a function here? E.g. `List (String \x Term)`? Or do we not know all the `attr`s?

NKL/Trace/Types.lean
```diff
@@ -0,0 +1,268 @@
...
+  | any    : Name -> TermType
+  | tuple  : List TermType -> TermType
+  | list   : List TermType -> TermType
+  | tensor : Dtype -> Shape -> TermType
+  deriving Repr, BEq
+
+mutual
+structure Builtin where
+  name : Name
+  type : TermType
```
seanmcl, line 84: We don't need the types of the arguments?

NKL/Trace/Types.lean
```diff
@@ -0,0 +1,268 @@
...
+  attr : String -> Except String Term
+  call : List Expr -> List (String × Expr) -> Except String Term
+
+inductive Term where
+  | object : Builtin -> Term
```
seanmcl, line 89: Can we call this `builtin` or rename `Builtin` to `Object`? Or explain why are there two different names?

NKL/Trace/Types.lean
```diff
@@ -0,0 +1,268 @@
...
+
+instance : Repr Term where
+  reprPrec b n := match b with
+    | .object obj => .text s!"object<{obj.name}>"
+    | .tuple l => .text s!"tuple<{l.length}>"
+    | .list l => .text s!"list<{l.length}>"
+    | .expr e ty  => reprPrec e n ++ ":" ++ reprPrec ty n
+
+def Term.beq : Term -> Term -> Bool
+  | .tuple l , .tuple r  => lst_eq l r
```
seanmcl, line 103: You can drop the first `=> list_eq l r` if you want.

NKL/Trace/Types.lean
```diff
@@ -0,0 +1,268 @@
...
+  | _, _ => false
+where
+  lst_eq : List Term -> List Term -> Bool
+  | [] , [] => true
+  | l :: ls, r :: rs => Term.beq l r && lst_eq ls rs
+  | _, _ => false
+
+instance : BEq Term where beq := Term.beq
+
+partial def Term.type : Term -> ErrorM TermType
```
seanmcl, line 115: Why `partial`?

NKL/Trace/Types.lean
```diff
@@ -0,0 +1,268 @@
...
+
+-- Run a trace with an empty initial environment
+def trace (m : TraceM a) : ErrorM a :=
+  match m.run { } with
+  | .ok x _ => return x
+  | .error s _ => throw s
+
+-- generate a fresh name using an existing name as a prefix
+def genName (name : Name := .anonymous) : TraceM Name := do
+  let s <- get
```
seanmcl, line 158:
```
let  s <- getModify fun n => n + 1
```

is marginally nicer than this 3 lines.

NKL/Trace/Types.lean
```diff
@@ -0,0 +1,268 @@
...
+  let n := s.fvn + 1
+  set { s with fvn := n }
+  return .num name n
+
+-- add a new binding to the local environment
+def extend (x : Name) (v : Term) : TraceM Unit :=
+  modify fun s => {s with env := s.env.insert x v}
```
seanmcl, line 165: Do you want to ensure the name isn't there already? Do we have scoping issues to worry about?

NKL/Trace/Types.lean
```diff
@@ -0,0 +1,268 @@
...
+
+-- lookup a name in the local environment
+def lookup? (name : Name) : TraceM (Option Term) := do
+  let s <- get
+  return s.env.find? name
+
+def lookup (name : Name) : TraceM Term := do
+  match (<- get).env.find? name with
```
seanmcl, line 173: This is just `lookup?`. May be marginally better above to inline it like you have here also.

NKL/Trace/Types.lean
```diff
@@ -0,0 +1,268 @@
...
+  | none => throw s!"{name} not found"
+  | some x => return x
+
+
+-- The Tracer monad is setup similar to the TraceM monad, but "one level up".
```
seanmcl, line 178: IMO the `--` comment syntax is kind of noisy for big comments.

NKL/Trace/Types.lean
```diff
@@ -0,0 +1,268 @@
...
+-- Built-ins in Tracer can operate over Terms and within the TraceM monad, hence
+-- they can query and modify the state of the inner monad.
```
seanmcl, line 180: This is subtle. It would be nice for posterity to give an intuition of why you can't have this in one monad. I know you showed me by flattening the types and showing the Lean knot you get into, but I wonder if there is a simple example (not really in lean per se) that would help the reader grok the issue.

NKL/Trace/Types.lean
```diff
@@ -0,0 +1,268 @@
...
+-- global environment as a convient way to represent global constants, such as
+-- `nki.language.psum`. Only the core tracing code needs to use the `Tracer`
+-- monad.
+
+-- The global environment also track the current source position for error
+-- reporting, and keeps the list of translated statements for the current
+-- kernel.
+
+structure Global where
+  name : Name
```
seanmcl, line 193: Why did we need the return type before but don't need it here?

NKL/Trace/Types.lean
```diff
@@ -0,0 +1,268 @@
...
+  attr : String -> TraceM Term
+  call : List Term -> List (String × Term) -> TraceM Term
+
+inductive Item where
+  | module : Name -> Item
+  | global : Global -> Item
+  | source : Python.Fun -> Item
```
seanmcl, line 200: Example?

NKL/Trace.lean
```diff
@@ -0,0 +1,9 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Paul Govereau
+-/
+import NKL.Trace.Types
+--import NKL.Trace.Basic
```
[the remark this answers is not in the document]
govereau: Ah, sure. I had just commented this as it appear in the next PR.

NKL/Trace/Types.lean
govereau: I am not sure, but I think instances like this will be visible to anyone that imports the module, regardless of scope.

NKL/Trace/Types.lean
govereau: I actually had that at one point, I can't remember why I changed it. Probably this needs a lot of reworking as we implement type inference for the tracing process.

NKL/Trace/Types.lean
govereau: I have the same thought here. Technically we don't know much about the attributes; they may even be properties which require computation. However, we could make some conventions about how we want to implement the builtins. I am inclined to leave it more general for now until we have some experience implementing the NKI builtins.

NKL/Trace/Types.lean
govereau: Well, a builtin may no be callable. An example is `nki.language.tile_size`, which is just a bunch of properties that return integers. I am note sure what `type` should be in this case? Perhaps `TermType` needs a function type, or even an object type? I don't know how far we need to go for tracing.

NKL/Trace/Types.lean
govereau: Oops, thanks. That is confusing.

NKL/Trace/Types.lean
govereau: I am being lazy and using `mapM`, which confuses the termination checker. It is easy to change if we need to.

NKL/Trace/Types.lean
govereau: I think we can shadow names, for instance:
```
x = 1
for x in range(10):
  print(x)
```
We, `enter` the for loop, and then `extend x`, which should hide the outer x. Maybe I should move `extend` to the `Tracer` monad so it is not usable by most code, helping to ensure proper use?

NKL/Trace/Types.lean
govereau: Hmm, I couldn't see a nice way to use this. Would it be:
```
  let old <- getModify fun s => {s with fvn := s.fvn + 1}
  return old.fvn + 1
```

NKL/Trace/Types.lean
govereau: Oh, yeah. nice catch.

NKL/Trace/Types.lean
govereau: Yes, agreed. I think the comment just kept getting bigger and bigger and I didn't convert. Will do.

NKL/Trace/Types.lean
govereau: I will expand the comments.

NKL/Trace/Types.lean
govereau: Probably we do need it here. I am just unsure exactly what we need to do. I think if we start implementing globals and the related type inference it will become clear.

NKL/Trace/Types.lean
govereau: I have extended the comments with examples.

NKL/Trace/Types.lean
seanmcl: Hmm, good point. Was thinking it could just be

```
do
  let s <- getModify fun s => { s with fvn := s.fvn + 1 })
  return .num name s.fvn
```

But that doesn't actually look much better than yours now that I write it out. Nevermind.

NKL/Trace/Types.lean
seanmcl: Right, but when the loop goes out of scope, the original value for `x` is not restored, is it? That was my concern at least, given that changes to the monad state persist.

seanmcl (approved):

PR #19, opened 2025-01-08
feat: basic tracing for Terms

govereau:
Basic tracing of Terms in the inner TraceM monad.

### DIFF
--- a/NKL/Trace/Basic.lean
+++ b/NKL/Trace/Basic.lean
@@ -1,0 +1,161 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Paul Govereau
+-/
+import NKL.KLR
+import NKL.Trace.Types
+import NKL.Trace.Tensor
+
+/-
+# Basic tracing facilities
+
+Basic tracing definitions only deal with Terms (not Python sources)
+-/
+
+namespace NKL.Trace
+open NKL.KLR
+
+-- Operators within index expressions
+
+def indexBinOp : String -> KLR.IndexExpr -> KLR.IndexExpr -> ErrorM KLR.IndexExpr
+  | "Add" ,      l,      r => return .add l r
+  | "Sub" ,      l,      r => return .add l r.neg
+  | "Mult", .int i,      e
+  | "Mult",      e, .int i => return .mul i e
+  | "Div" ,      e, .int i => return .floor e i
+  | "Mod" ,      e, .int i => return .mod e i
+  | _, _, _ => throw "invalid index expression"
+
+def indexUnOp : String -> KLR.IndexExpr -> ErrorM KLR.IndexExpr
+  | "USub", e => return .neg e
+  | _, _ => throw "invalid index expresssion"
+
+-- Truthiness of Terms following Python
+
+def Term.isTrue : Term -> TraceM Bool
+  | .object _ => return true
+  | .tuple [] => return false
+  | .tuple _  => return true
+  | .list []  => return false
+  | .list _   => return true
+  | .expr (.const c) _ => return c.isTrue
+  | .expr _          _ => throw "non-constant expression"
+
+def Term.isFalse (t : Term) : TraceM Bool :=
+  return not (<- t.isTrue)
+
+-- Following Python semantics, boolean operators return
+-- the first value that is convertible to True or False
+
+def boolOp (op : String) (es : List Term) : TraceM Term := do
+  bop (<- bopFn op) es
+where
+  bop fn : List Term -> TraceM Term
+    | []  => throw "invalid expression"
+    | [x] => return x
+    | x :: xs => do if (<- fn x) then return x else bop fn xs
+  bopFn : String -> TraceM (Term -> TraceM Bool)
+    | "Or"  => return Term.isTrue
+    | "And" => return Term.isFalse
+    | s     => throw s!"unsupported boolean operator {s}"
+
+-- Binary Operators
+
+-- Multiply a sequence (tuple, list, string) by a scalar
+-- It is tempting to use Const.toInt here, but that would be
+-- more permissive than Python. The only allowed cases are:
+--   [1,2] * 2     => [1,2,1,2]
+--   [1,2] * 0     => []
+--   [1,2] * -10   => []
+--   [1,2] * True  => [1,2]
+--   [1,2] * False => []
+
+private def mulseq (l : List a) : Const -> TraceM (List a)
+  | .bool false => return []
+  | .bool true  => return l
+  | .int i      => return List.flatten $ List.replicate i.toNat l
+  | _           => throw "invalid multiply"
+
+-- Binary operators on constants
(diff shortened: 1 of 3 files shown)


NKL/Trace/Basic.lean
```diff
@@ -0,0 +1,158 @@
...
+    | [x] => return x
+    | x :: xs => do if (<- fn x) then return x else bop fn xs
+  bopFn : String -> TraceM (Term -> TraceM Bool)
+    | "Or"  => return Term.isTrue
+    | "And" => return Term.isFalse
+    | s     => throw s!"unsupported boolean operator {s}"
+
+-- Binary Operators
+
+-- Multiply a sequence (tuple, list, string) by a scalar
```
seanmcl, line 65: This is so nice to have written down.

NKL/Trace/Basic.lean
```diff
@@ -0,0 +1,158 @@
...
+--   [1,2] * 2     => [1,2,1,2]
+--   [1,2] * 0     => []
+--   [1,2] * -10   => []
+--   [1,2] * True  => [1,2]
+--   [1,2] * False => []
+
+private def mulseq (l : List a) : Const -> TraceM (List a)
+  | .bool false => return []
+  | .bool true  => return l
+  | .int i      => return (List.flatten $ List.replicate i.toNat l)
```
seanmcl, line 77: super-nit: return binds very loosely and doesn't need braces.

NKL/Trace/Basic.lean
```diff
@@ -0,0 +1,158 @@
...
+
+-- Unary operators
+def unOp : String -> Term -> TraceM Term
+  | op   , .expr (.tensor t) _ => tensor_op op t
+  | "Not", t                   => return .expr (.const $ .bool (<- t.isFalse)) .bool
+  | op, _ => throw s!"unimp {op}"
+
+-- Comparison operators
+def cmpOp : String -> Term -> Term -> TraceM Bool
+  | "Eq", .expr l _, .expr r _ => return l == r
```
seanmcl, line 122: Why does python `==` match `BEq` here?

NKL/Trace/Basic.lean
```diff
@@ -0,0 +1,158 @@
...
+  | .global g, id => return .term (<- g.attr id)
+  | .source _, id => throw s!"unsupported attribute {id}"
+  | .term   t, id => return .term (<- t.attr id)
+
+def Term.call (f : Term)
+              (args : List Expr)
+              (kws : List (String × Expr)) : TraceM Term := do
+  match f with
+  | .object o => o.call args kws
+  | .tuple _  => throw "tuple is not a callable type"
```
seanmcl, line 156:
```
In [144]: [1,2,3][0]
Out[144]: 1

In [145]: (1, 2, 3)[0]
Out[145]: 1
```

NKL/Trace/Basic.lean
```diff
@@ -0,0 +1,158 @@
...
+  | .list _   => throw "list is not a callable type"
+  | .expr f _ => return .expr (.call f args kws) (.any "?".toName)
```
seanmcl, line 158: I don't understand these `"?"` names.

NKL/Trace/Tensor.lean
```diff
@@ -0,0 +1,43 @@
...
+private def broadcast (t : Tensor) (c : Const) : Expr :=
+  let args := t.shape.map fun i => Expr.const (.int i)
+  let args := .const c :: args
+  .call (.var "broadcast") args []
+
+def tensor_scalar (op : String) (t : Tensor) (c : Const) : TraceM Term :=
+  return tensor_call op [ .tensor t, broadcast t c]
+
+def scalar_tensor (op : String) (c : Const) (t : Tensor) : TraceM Term :=
+  return tensor_call op [ .tensor t, broadcast t c]
```
seanmcl, line 43: I guess all tensor/scalar ops are commutative? If so maybe we don't need both of these.

NKL/Trace/Basic.lean
```diff
@@ -0,0 +1,158 @@
...
+Basic tracing definitions only deal with Terms (not Python sources)
+-/
+
+namespace NKL.Trace
+open NKL.KLR
+
+-- Operators within index expressions
+
+def indexBinOp : String -> KLR.IndexExpr -> KLR.IndexExpr -> ErrorM KLR.IndexExpr
+  | "Add" ,      l,      r => return .add l r
```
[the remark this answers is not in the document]
govereau: These are from https://docs.python.org/3/library/ast.html
I think we should convert these to an enum earlier in the process. It is just a todo item for me to go and check if all the versions of python we need to support actually have the same set of operators.

NKL/Trace/Basic.lean
govereau: Yes, thank you. I prefer less braces when possible.

NKL/Trace/Basic.lean
govereau: Oh! I think it doesn't, nice catch.
We should not have any variables of function calls at this point, but tensors...
It looks like `==` on `nki.tensor` is a pointer equality on the objects, and we have something more permissive (just checking the dtype and shape are the same).

Hmm, I am not sure you can even compare tensors in NKI? I need to investigate this. Regardless, I will break out the equality and generate errors for the cases I am not sure about.

NKL/Trace/Basic.lean
govereau: Sorry, not understanding your comment.

NKL/Trace/Basic.lean
govereau: Sorry about that. It just means "I have no idea, we need type inference" :D (and probably more types).

NKL/Trace/Basic.lean
seanmcl: Isn't this an example of a tuple being callable? Oh, I think I'm confusing "indexable" and "callable".

NKL/Trace/Tensor.lean
govereau: I don't know. We have subtraction and division, but I don't know if you are actually allowed to write those expressions? That is, the AST can represent:
```
1 - tensor   # same as tensor_negate(tensor) + 1 ?
2 / tensor   # same as tensor_invert(tensor) * 2 ?
``` 
but I don't know if this is actually allowed in NKI. Even if not allowed, maybe it should be?

NKL/Trace/Basic.lean
govereau: I see. Technically you are correct, because in python,
```
a[0]   ==> a.__subscript__(0)
```
However, the python parser gives you a subscript node, which is handled as its own thing. If we were being really careful, we would have an attribute called "__subscript__" on tuples, which returned a builtin function, etc... not sure if it is worth it.

seanmcl: yeah that's fine, I just missed that it uses different brackets

NKL/Trace/Basic.lean
govereau: Just to complete the loop here. I am going to merge this with some comments, and create a follow-up PR.

PR #20, opened 2025-01-09
Tracing for python source functions

govereau:
This patch adds basic tracing for user python functions. The main code
is in Python.lean, and depends on definitions in Basic.lean and
NKI.lean, which are incomplete. As more primitives are implemented,
more user kernels will be supported.

Note to reviewers: it may be helpful to check the tests in `interop/tests/test_basic.py`.

### DIFF
--- a/NKL/Trace.lean
+++ b/NKL/Trace.lean
@@ -3,7 +3,18 @@ Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Paul Govereau
 -/
+import NKL.KLR
+import NKL.Python
 import NKL.Trace.Types
 import NKL.Trace.Basic
 import NKL.Trace.Builtin
---import NKL.Trace.Python
+import NKL.Trace.Python
+import NKL.Trace.NKI
+
+namespace NKL.Trace
+
+def runNKIKernel (k : NKL.Python.Kernel) : Except String (List NKL.KLR.Stmt) :=
+  tracer ⟨ .ofList NKIEnv, #[] ⟩ do
+    traceKernel k
+    let g <- get
+    return g.body.toList
--- a/NKL/Trace/Basic.lean
+++ b/NKL/Trace/Basic.lean
@@ -118,11 +118,8 @@ def unOp : String -> Term -> TraceM Term
   | op, _ => throw s!"unimp {op}"
 
 -- Comparison operators
--- TODO: need to think about comparison of tensors, in NKI this is object equality,
--- but I suspect this doesn't make sense and may be a source of bugs.
 
 def cmpOp : String -> Term -> Term -> TraceM Bool
-  | "Eq", .expr l _, .expr r _ => return l == r
   | s, l, r => throw s!"unsupported comparison operator {s} {repr l} {repr r}"
 
 def compare : Term -> List String -> List Term -> TraceM Term
--- a/NKL/Trace/NKI.lean
+++ b/NKL/Trace/NKI.lean
@@ -1,0 +1,49 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Paul Govereau
+-/
+import NKL.KLR
+import NKL.Trace.Types
+import NKL.Trace.Builtin
+
+/-
+# NKI built-ins
+
+This module defines the builtin constants used by tracing for NKI kernels.
+-/
+namespace NKL.Trace
+open NKL.KLR
+
+private def module (s : String) : Name × Item :=
+  let name := s.toName
+  (name, .module name)
+
+private def const_var (s : String) : Name × Item :=
+  let name := s.toName
+  (name, .term (.expr (.var s) (.any name)))
+
+/-
+Note: this object contains a bunch of architecture parameters that
+need to be set according to which HW we are compiling for.
+TODO: figure out the mechanism for this.
+-/
+def tile_size : Global :=
+  let name := "nki.langauge.tile_size".toName
+  { name := name
+  , attr := attrs
+  , call := uncallable name
+  }
+where
+  attrs : GlobalAttr
+  | "pmax" => return .expr (.const $ .int 128) .int
+  | a => throw s!"unsupported attribute {a}"
+
+def NKIEnv : List (Name × Item) :=
+  [ module "nki"
+  , module "nki.language"
+  , const_var "nki.language.add"
+  , const_var "nki.language.load"
+  , const_var "nki.language.store"
+  , ("nki.language.tile_size".toName, .global tile_size)
+  ]
--- a/NKL/Trace/Python.lean
+++ b/NKL/Trace/Python.lean
@@ -1,0 +1,264 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Paul Govereau
+-/
+import Lean
+import NKL.KLR
+import NKL.Python
+import NKL.Trace.Types
+import NKL.Trace.Basic
+
+namespace NKL.Trace
+open NKL.Python
+
+def const : Const -> ErrorM Term
+  | .none     => return .expr (.const $ .none)     .none
+  | .bool b   => return .expr (.const $ .bool b)   .bool
+  | .int i    => return .expr (.const $ .int i)    .int
+  | .float f  => return .expr (.const $ .float f)  .float
+  | .string s => return .expr (.const $ .string s) .string
+  | .ellipsis => throw "unsupported use of ellipsis"
+
+/-
+Evaluating index expressions.
+
+An index expression occurs only within a subscript expression. For example, in
+the expression:
+
+  t[1,1:10,None,x+9]
+
+all of 1, 1:10, None, and x+9 are indexes. Note None may also be written as
+np.newaxis. Also, a None or a slice (or ellipsis) may only occur at the
+outer-most level of an index: if you write, e.g.
+
+  t[x+None]
+
+then the None is interpreted as an integer and not as a new axis. If you write,
+
+  t[(1:2) + 3]
+  t[... * 8]
+
+these are syntax errors in python.
+-/
+
+mutual
+-- top-level index expressions: None (a.k.a. np.newaxis) or IndexExpr
+def indexExpr? : Option Expr -> Tracer (Option KLR.IndexExpr)
+  | none   => return none
+  | some (.exprPos (.const .none) _) => return none
+  | some e => indexExpr e
+
+-- general sub-expressions
+def indexExpr : Expr -> Tracer KLR.IndexExpr
+  | .exprPos e' p => withPos p (indexExpr' e')
+
+def indexExpr' : Expr' -> Tracer KLR.IndexExpr
+  | .const (.int i) => return .int i
+  | .name id _      => return .var id
+  | .binOp op l r   => return <- indexBinOp op (<- indexExpr l) (<- indexExpr r)
+  | .unaryOp op e   => return <- indexUnOp op (<- indexExpr e)
+  | _ => throw "invalid index expression"
+
+-- top-level index: slice, ellipsis, or indexExpr?
+-- TODO: get rid of ...
+def index : Expr -> Tracer KLR.Index
+  | .exprPos (.const .ellipsis) _ => return .ellipsis
+  | .exprPos (.slice l u s) p => withPos p do
+      return (.slice (<- indexExpr? l) (<- indexExpr? u) (<- indexExpr? s))
+  | e => return (.coord (<- indexExpr? e))
+end
+
+-- Note, a list index can be negative, which means index from end of list.
+def list_access (l : List Term) : List KLR.Index -> TraceM Term
+  | [.coord (some (.int i))] => do
+      let i := if i < 0 then l.length + i else i
+      if i < 0 then throw "index out of bounds"
+      let n := i.toNat
+      if h:l.length > n then return l.get (Fin.mk n h)
+      else throw "index out of bounds"
+  |_ => throw "unsupported __subscript__"
+
+def access : Term -> List KLR.Index -> TraceM Term
+  | .object _, _  => throw "builtin object __subscript__ not supported"
+  | .tuple l, ix
+  | .list l, ix   => list_access l ix
+  | .expr e _, ix => return .expr (.access e ix) (.any "?".toName)
+
+
+mutual
+partial def expr : Expr -> Tracer Item
+  | .exprPos e' p => withPos p (expr' e')
+
+partial def term (e : Expr) : Tracer Term := do
+  match (<- expr e) with
+  | .module n   => return .expr (.var n.toString) (.any "?".toName)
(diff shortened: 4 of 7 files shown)


NKL/Trace.lean
```diff
@@ -3,7 +3,18 @@ Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
...
 import NKL.Trace.Builtin
---import NKL.Trace.Python
+import NKL.Trace.Python
+import NKL.Trace.NKI
+
+namespace NKL.Trace
+
+def runNKIKernel (k : NKL.Python.Kernel) : Except String (List NKL.KLR.Stmt) :=
+  tracer ⟨ .ofList nki_env, #[] ⟩ do
+    traceKernel k
```
seanmcl, line 18: This is a kind of annoying aspect of Lean imo. Using the namespace mechanism, I now have no idea where to look for `traceKernel` without UI support. I wonder if we just get used to it, or have some conventions around files vs namespace organization.

NKL/Trace/NKI.lean
```diff
@@ -0,0 +1,32 @@
...
+
+private def module (s : String) : Name × Item :=
+  let name := s.toName
+  (name, .module name)
+
+private def const_var (s : String) : Name × Item :=
+  let name := s.toName
+  (name, .term (.expr (.var s) (.any name)))
+
+def nki_env : List (Name × Item) :=
```
seanmcl, line 26: Why snake_case here?

NKL/Trace/Python.lean
```diff
@@ -0,0 +1,234 @@
...
+  | .string s => return .expr (.const $ .string s) .string
+  | .ellipsis => throw "unsupported use of ellipsis"
+
+mutual
+def indexExpr : Expr -> Tracer KLR.IndexExpr
+  | .exprPos e' p => withPos p (indexExpr' e')
+
+def indexExpr' : Expr' -> Tracer KLR.IndexExpr
+  | .const (.int i)   => return .int i
+  | .const c          => throw s!"invalid constant {repr c} in index expression"
```
seanmcl, line 29: Wouldn't `.none` be `np.newaxis`? That's used a lot in kernels.

NKL/Trace/Python.lean
```diff
@@ -0,0 +1,234 @@
...
+def index : Expr -> Tracer KLR.Index
+  | .exprPos (.const .ellipsis) _ => return .ellipsis
+  | .exprPos (.slice l u s) p => withPos p do
+      return (.slice (<- indexExpr? l) (<- indexExpr? u) (<- indexExpr? s))
+  | e => return (.coord (<- indexExpr? e))
+end
+
+def list_access (l : List Term) : List KLR.Index -> TraceM Term
+  | [.coord (some (.int i))] =>
+      let n := i.toNat
```
seanmcl, line 49: This ignores Python negative integer interpretations. Does NKI not allow things like 
```
In [146]: [1,2,3][-1]
Out[146]: 3
```

NKL/Trace/Python.lean
```diff
@@ -0,0 +1,234 @@
...
+--   def add(x,y): pass
+-- If we have an internal definition, we will use this over anything found
+-- during parsing.
+
+private def globals (k : Kernel) : Tracer Unit := do
+  let s <- get
+  for (n, f) in k.funcs do
+    let n := n.toName
+    if not (s.env.contains n) then
+      extend_global n (.source f)
```
seanmcl, line 214: I can't find this definition. This updates the TraceM inner monad? Maybe I'll take a break and catch up when you merge some of the earlier PRs.

NKL/Trace/Python.lean
```diff
@@ -0,0 +1,234 @@
...
+                          : Tracer Unit := do
+  let args <- bind_args f args kwargs
+  let args <- args.mapM fun (x,e) => return (x, e)
+  withSrc f.source $ enterFun $ do
+    args.forM fun (x,e) => do extend x.toName e
+    f.body.forM stmt
+
+end
+
+-- Evaluate each global in the current environment, skipping any globals that
```
seanmcl, line 202: Why would we skip globals already defined? What makes the old one better than the new one?

NKL/Trace/Python.lean
```diff
@@ -0,0 +1,234 @@
...
+  return argmap
+
+-- For a function call, first evaluate the argument in the current environment.
+-- Then enter a new environment and evaluate the function statements.
+partial def function_call (f : Fun)
+                          (args : List Term)
+                          (kwargs : List (String × Term))
+                          : Tracer Unit := do
+  let args <- bind_args f args kwargs
+  let args <- args.mapM fun (x,e) => return (x, e)
```
seanmcl, line 195: I imagine that `xs.mapM (fun x => return x) == return xs`

NKL/Trace/Python.lean
```diff
@@ -0,0 +1,234 @@
...
+      stmt' (.assign [x] (.exprPos (.binOp op x e) (<- getPos)))
+  | .annAssign _ _ .none => return ()
+  | .annAssign x _ (.some e) => stmt' (.assign [x] e)
+  | _s => throw "not yet implemented" --s!"unimp {repr s}"
+
+-- Bind positional and keyword arguments to a Python function.
+-- Note: default arguments should be evaluated in the global environment,
+-- however we know that each source function begins with an empty local
+-- environment, so it is OK to evaluate the default arguments in the
+-- functions initial environment.
```
seanmcl, line 163: Nice comment. Nit: "function's"

NKL/Trace/Python.lean
```diff
@@ -0,0 +1,234 @@
...
+  | .augAssign x op e => do
+      stmt' (.assign [x] (.exprPos (.binOp op x e) (<- getPos)))
+  | .annAssign _ _ .none => return ()
+  | .annAssign x _ (.some e) => stmt' (.assign [x] e)
+  | _s => throw "not yet implemented" --s!"unimp {repr s}"
+
+-- Bind positional and keyword arguments to a Python function.
+-- Note: default arguments should be evaluated in the global environment,
+-- however we know that each source function begins with an empty local
+-- environment, so it is OK to evaluate the default arguments in the
```
seanmcl, line 162: > we know that each source function begins with an empty local environment

How?

NKL/Trace/Python.lean
```diff
@@ -0,0 +1,234 @@
...
+                      : Tracer (List (String × Term)) := do
+  if f.args.vararg != none || f.args.kwarg != none then
+    throw "var args not supported"
+  if args.length < f.args.posonlyargs.length then
+    throw "not enough arguments"
+  let dflts := f.args.all_defaults
+  let names := f.args.names
+  if args.length + kwargs.length > names.length then
+    throw "too many arguments supplied (varargs not supported)"
+  let argmap <- f.args.names.enum.mapM fun (i,x) => do
```
seanmcl, line 177: Nice to know how python gets argument values now!

NKL/Trace/Python.lean
```diff
@@ -0,0 +1,234 @@
...
+      match <- term e with
+      | .expr e _ => add_stmt (.expr e)
+      | _ => return ()  -- effects are done, can be removed from KLR
+  | .assert e => do
+      let t <- term e
+      if (<- t.isFalse) then throw "assertion failed"
+  | .assign xs e => assign xs e
+  | .augAssign x op e => do
+      stmt' (.assign [x] (.exprPos (.binOp op x e) (<- getPos)))
+  | .annAssign _ _ .none => return ()
```
seanmcl, line 155: What is `annAssign` again?

NKL/Trace/Python.lean
```diff
@@ -0,0 +1,234 @@
...
+-- When looking for a variable we rely on the store attribute
+-- from the Python parser to check if it is a defining use.
+partial def var : Expr -> Tracer String
+  | .exprPos (.name id .store) _ => return id
+  | _ => throw "expecting variable"
+
+-- When we perform an assignment, we will either add to the environment
+-- the term found on the RHS, or the variable itself. The latter case
+-- allows us to lookup and find the variable without substituting
+-- it's definition.
```
seanmcl, line 129: its

NKL/Trace/Python.lean
```diff
@@ -0,0 +1,234 @@
...
+partial def assign (xs : List Expr) (e : Expr) : Tracer Unit := do
+  let xs <- xs.mapM var
+  let e <- term e
+  match e with
+  | .expr (.const _) _ => xs.forM fun x => extend x.toName e
+  | .expr e ty => xs.forM fun x => do
+      extend x.toName (.expr (.var xs[0]!) ty)
+      add_stmt (KLR.Stmt.assign x e)
+  | t => xs.forM fun x => do extend x.toName t
```
seanmcl, line 138: Think you can remove the `do` here.

NKL/Trace/Python.lean
```diff
@@ -0,0 +1,234 @@
...
+  | .unaryOp op e     => return <- indexUnOp op (<- indexExpr e)
+  | _ => throw "invalid index expression"
+
+def indexExpr? : Option Expr -> Tracer (Option KLR.IndexExpr)
+  | none   => return none
+  | some (.exprPos (.const .none) _) => return none
+  | some e => indexExpr e
+
+def index : Expr -> Tracer KLR.Index
+  | .exprPos (.const .ellipsis) _ => return .ellipsis
```
seanmcl, line 41: Maybe we should make more of an effort to get rid of ... if possible.

NKL/Trace/NKI.lean
govereau: Oops, will fix.

NKL/Trace/Python.lean
govereau: That's a bug. Will fix.

NKL/Trace/Python.lean
govereau: I will add some comments. Basically, I don't allow users to override the builtins. This is convenient because the NKI interface files have empty definitions which are picked up by the Python code.

NKL/Trace/Python.lean
govereau: Thanks, fixed.

NKL/Trace/Python.lean
govereau: Fixed.

NKL/Trace/Python.lean
govereau: I will change this. There is a simple refactoring which makes the code more clear.

NKL/Trace/Python.lean
govereau: Thanks.

NKL/Trace/Python.lean
govereau: Yeah, agree. I will add a comment and follow-up with another PR. I may need your help to figure this out.

NKL/Trace.lean
seanmcl: This is my gut feeling, but I may be missing some bigger picture here. I just feel like all module systems I'm aware of that don't respect the file system are harder to use/find things.

NKL/Trace/Basic.lean
```diff
@@ -122,7 +122,8 @@ def unOp : String -> Term -> TraceM Term
 -- but I suspect this doesn't make sense and may be a source of bugs.
 
 def cmpOp : String -> Term -> Term -> TraceM Bool
-  | "Eq", .expr l _, .expr r _ => return l == r
+  | "Eq", l, r => return l == r
```
seanmcl, line 125: I still don't grok why the Lean `Eq` class is related to Python's `==`. At least could we have a comment about this?

NKL/Trace/NKI.lean
```diff
@@ -0,0 +1,44 @@
...
+
+def tile_size : Global :=
+  let name := "nki.langauge.tile_size".toName
+  { name := name
+  , attr := attrs
+  , call := uncallable name
+  }
+where
+  attrs : GlobalAttr
+  | "pmax" => return .expr (.const $ .int 128) .int
```
seanmcl, line 34: Hmm, this seems brittle wrt new generations. I wonder if we should have a global config file or something.

NKL/Trace/Python.lean
```diff
@@ -0,0 +1,264 @@
...
+import Lean
+import NKL.KLR
+import NKL.Python
+import NKL.Trace.Types
+import NKL.Trace.Basic
+
+namespace NKL.Trace
+open NKL.Python
+
+def const : Const -> TraceM Term
```
seanmcl, line 15: Is this type overly complicated? Seems this could just be `Except String Term`? Maybe doesn't matter. I just kept waiting to see when you required some state.

seanmcl (commented):
Still some unresolved comments from last time. If you want to punt them to the future just add some TODO comments and I'll approve.

NKL/Trace/Basic.lean
govereau: I think it is incorrect, I will remove it for now and take a pass at doing all the operators in full detail.

NKL/Trace/NKI.lean
govereau: I think this is also an issue in the official NKI. There needs to be some way to say which HW you are compiling for and to query which HW you are compiling for. I don't think there is a proposal for this yet. I will add comments warning that we have work to do here.

NKL/Trace/Python.lean
govereau: Good call.
Also, I think I used `ErrorM` and you used `Err`, so I should search and replace this, but seems like a separate PR.

NKL/Trace/Python.lean
```diff
@@ -0,0 +1,264 @@
...
+
+  t[(1:2) + 3]
+  t[... * 8]
+
+these are syntax errors in python.
+-/
+
+mutual
+-- top-level index expressions (None or IndexExpr)
+def indexExpr? : Option Expr -> Tracer (Option KLR.IndexExpr)
```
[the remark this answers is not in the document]
govereau: I will improve the commment.

interop/test/test_basic.py
```diff
@@ -0,0 +1,93 @@
...
+
+def expr_subscript(t):
+  t[1]
+  t[1,2,3]
+  t[1:2:3]
+  t[1:2]
+  t[1:]
+  t[1::]
+  t[1::2]
+  t[1:2:None]
```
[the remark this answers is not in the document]
govereau: Actually, I dealt with that. Regardless of how the user imports we will always get `numpy.newaxis`. So, it should be easy. I was thinking we need to have a Numpy file of builtins in the Trace directory, and this should just use TensorLib? (newaxis would be in there, defined as None?)

interop/test/test_basic.py
```diff
@@ -0,0 +1,93 @@
+import numpy as np
```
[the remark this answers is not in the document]
govereau: Yes, I couldn't figure out how to get the output back to python easily. Maybe we need a lean based testing framework?

NKL/Trace/Python.lean
govereau: That is handled at the top-level. I think None can only occur at the top-level of an index?
e.g. `t[None,...]`, but not `t[3+None,...]` (although that last one probably works in python?

NKL/Trace/Python.lean
```diff
@@ -0,0 +1,264 @@
...
+  | none   => return none
```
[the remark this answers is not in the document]
govereau: Happy to switch.

seanmcl (approved):

PR #21, opened 2025-01-21
A simple pretty-printer for KLR

govereau:
This change adds two related things: a pretty-printer for KLR terms, and tensor names. Tensor names make the pretty printing nicer, but have a second purpose. By naming all of the tensors, we can scan a KLR kernel to collect up all of the input, output, and intermediate tensors that will be needed to run the kernel. For argument tensors, the generated tensor names are changed to the argument variable names; this is just for readability.

### DIFF
--- a/NKL/KLR/Basic.lean
+++ b/NKL/KLR/Basic.lean
@@ -14,16 +14,22 @@ portable format, a.k.a. Kernel Language Representation (KLR).
 
 namespace NKL.KLR
 
--- TODO switch to tensor lib
+-- TODO switch to TensorLib's version of these types
 --export TensorLib (Tensor Dtype Shape)
--- Mostly, NKL deals with empty tensors, so just check dtype and shape
--- TODO: talk to Sean about a more general BEq for Tensor
---instance : BEq Tensor where
---  beq t₁ t₂ := t₁.dtype == t₂.dtype && t₁.shape == t₂.shape
 
 abbrev Dtype := String
 abbrev Shape := List Int
-structure Tensor where
+
+/-
+A TensorName is essentially a typed variable, where the type
+must be a tensor type. When we flush out Typ below we may replace
+this with `Expr.var name (Typ.tensor dtype shape)`. For now, this
+only refers to dynamic tensors, or compile-time tensors, not
+trace-time tensors.
+-/
+
+structure TensorName where
+  name  : String
   dtype : Dtype
   shape : Shape
   deriving Repr, BEq
@@ -71,7 +77,7 @@ def toInt : Const -> Except String Int
 
 end Const
 
--- This correspondes to the "Quasi-Affine Expressions" in Neuron.
+-- This corresponds to the "Quasi-Affine Expressions" in Neuron.
 -- Note, `floor` is the usual integer division.
 inductive IndexExpr where
   | var (name : String)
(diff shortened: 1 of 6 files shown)


NKL/KLR/Basic.lean
```diff
@@ -24,6 +24,7 @@ namespace NKL.KLR
 abbrev Dtype := String
 abbrev Shape := List Int
 structure Tensor where
+  name  : String
```
seanmcl, line 27: Seems weird to require tensors to have names. I see why it's convenient here. Maybe we change the type to `NamedTensor` or `TensorName` to indicate it's not really a tensor at all?

NKL/KLR/Encode.lean
```diff
@@ -276,7 +276,7 @@ where
 partial def decExpr : DecodeM Expr := do
   match (<- next) with
   | 0x30 => return .var (<- decString)
-  | 0x31 => return .tensor $ .mk (<- decString) (<- decList decInt)
+  | 0x31 => return .tensor $ .mk (<- decString) (<- decString) (<- decList decInt)
```
seanmcl, line 279: This is the kind of thing I'm worried about in a hand-rolled serializer.

NKL/KLR/Pretty.lean
```diff
@@ -0,0 +1,84 @@
...
+
+instance : ToFormat Tensor    where format := ppTensor
+instance : ToFormat Const     where format := ppConst
+instance : ToFormat IndexExpr where format := ppIndexExpr 0
+instance : ToFormat Index     where format := ppIndex
+instance : ToFormat Expr      where format := ppExpr
+instance : ToFormat Stmt      where format := ppStmt
+
+@[default_instance]
+instance [ToFormat a]: ToString a where
```
seanmcl, line 83: Not crazy about this. Seems like formatting and to-stringing are different things. I'd be more sympathetic in the other direction.

NKL/Trace/Python.lean
```diff
@@ -208,16 +210,21 @@ partial def bind_args (f : Fun)
...
 
 -- For a function call, first evaluate the argument in the current environment.
 -- Then enter a new environment and evaluate the function statements.
 partial def function_call (f : Fun)
                           (args : List Term)
                           (kwargs : List (String × Term))
                           : Tracer Unit := do
-  let args <- bind_args f args kwargs
-  --let args <- args.mapM fun (x,e) => return (x, e)
+  let args <- bind_args f args kwargs true
```
seanmcl, line 227: Could we please name boolean arguments?

seanmcl (approved):

NKL/KLR/Basic.lean
govereau: Good point, Maybe `TensorName`.
Now I am wondering why I can translate kernels without any "real" tensors. Probably in current NKI there are no "real" tensors, but we may need them for advanced indexing.

NKL/KLR/Encode.lean
govereau: Yup, I think we should pick something and switch to it. Protobuf seems like the obvious choice.

NKL/KLR/Pretty.lean
govereau: That's fair, I will delete.

NKL/Trace/Python.lean
govereau: Good call.

PR #22, opened 2025-01-21
feat: add assignments to tuples and lists

govereau:
This change adds support for assignment like: `x, y = z`. Python calls this "unpacking", and you can convert between lists and tuples as long as they are the right length.

### DIFF
--- a/NKL/Trace/Python.lean
+++ b/NKL/Trace/Python.lean
@@ -21,7 +21,7 @@ def const : Const -> ErrorM Term
   | .ellipsis => throw "unsupported use of ellipsis"
 
 /-
-Evaluating index expressions.
+# Evaluating index expressions
 
 An index expression occurs only within a subscript expression. For example, in
 the expression:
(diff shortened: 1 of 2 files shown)


NKL/Trace/Python.lean
```diff
@@ -146,39 +253,17 @@ partial def expr' : Expr' -> Tracer Item
...
   | .stmtPos s' p => withPos p (stmt' s')
 
 partial def stmt' : Stmt' -> Tracer Unit
-  | .expr (.exprPos (.const _) _) => return ()
   | .expr e => do
-      match <- term e with
-      | .expr e _ => add_stmt (.expr e)
-      | _ => return ()  -- effects are done, can be removed from KLR
+      let t <- term e
+      let _ <- RValue t
```
govereau, line 262: I am not sure if this is clear. But my thinking is that if you have an expression as a statement, then we can use the `RValue` conversion to figure out if it "has side-effects". If it does, the `RValue` translation will insert an assignment to a fresh variable, otherwise we just discard the expression (e.g. doc strings, etc.)

NKL/Trace/Python.lean
```diff
@@ -85,6 +85,113 @@ def access : Term -> List KLR.Index -> TraceM Term
...
+  | [], [] => return ()
+  | [], _  => throw "not enough values to unpack"
+  | _, []  => throw "too many values to unpack"
+  | x::xs, t::ts => do
+      assignTerm x t;
+      assignList xs ts
+
+-- Top-level assignment handling
+-- e.g. x1 = x2 = e
+def assign (xs : List Term) (e : Term) : Tracer Unit := do
```
seanmcl, line 187: Pretty.

seanmcl (approved):

PR #23, opened 2025-01-27
refactor: make some partial functions total

govereau:
This patch cleans up a few cases of partial definitions, making them total. There is also a small amount of refactoring and renaming to be consistent with TensorLib.

### DIFF
--- a/NKL/Trace/Python.lean
+++ b/NKL/Trace/Python.lean
@@ -12,7 +12,7 @@ import NKL.Trace.Basic
 namespace NKL.Trace
 open NKL.Python
 
-def const : Const -> ErrorM Term
+def const : Const -> Err Term
   | .none     => return .expr (.const $ .none)     .none
   | .bool b   => return .expr (.const $ .bool b)   .bool
   | .int i    => return .expr (.const $ .int i)    .int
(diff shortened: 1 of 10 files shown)


NKL/Trace/Python.lean
```diff
@@ -132,21 +132,21 @@ dead-code elimination for simple assignments.
...
 -- Convert an expression in assignment context (an L-Value).
-partial def LValue : Expr -> Tracer Term
+def LValue : Expr -> Tracer Term
   | .exprPos e' p => withPos p (lval e')
 where
   lval : Expr' -> Tracer Term
   | .name id .store => return .expr (.var id) (.any "?".toName)
-  | .tuple l .store => return .tuple (<- l.mapM LValue)
-  | .list  l .store => return .list  (<- l.mapM LValue)
+  | .tuple l .store => return .tuple (<- LValue ▷ l)
```
seanmcl, line 140: Not sure how I feel about non-fixed-width unicode characters. Will suspend judgment. Maybe it'll be great.

NKL/Trace/Python.lean
```diff
@@ -197,45 +197,37 @@ mutual
...
-  | .expr e _   => return e
+partial def klr (e : Expr) : Tracer KLR.Expr :=
+  return <- (<- term e).toKLR
 
 partial def integer (e : Expr) : Tracer Int := do
-  match (<- term e) with
-  | .expr (.const c) _ => return (<- c.toInt)
-  | _ => throw "invalid tensor dimension"
+  match <- klr e with
+    | .const c => return (<- c.toInt)
```
seanmcl, line 211: indent seems off

NKL/Util.lean
```diff
@@ -0,0 +1,47 @@
...
+
+We can capture this trick in a notation. Note we need to use a notation and not
+a definition because the proof object `x∈l` needs to be available to the
+termination proof tactics, in the scope of the original function.
+
+Writing, `List.mapM f l`, as `f ▷ l` doesn't break the termination proof.
+Note: ▷ is typed as \rhd
+-/
+notation f "▷" l =>
+  List.mapM (fun ⟨ x, _ ⟩ => f x) (List.attach l)
```
seanmcl, line 47: I don't understand why this works. You're just dropping the proof term you're attaching?

NKL/Util.lean
```diff
@@ -0,0 +1,47 @@
+/-
```
seanmcl, line 1: Nice to capture all this Lean messiness here in one well-documented file.

seanmcl (approved):

NKL/Trace/Python.lean
govereau: Same, this is my first time using a unicode character.

NKL/Util.lean
govereau: I got this at the FRO office hours. The way I understand it is that, even though the proof is unnamed, it still exists in the context where the termination tactics run, so it is found and used.

PR #25, opened 2025-02-04
[WIP] subprocess version of python/lean communication

seanmcl:
Had to comment out the `assert` statements which aren't supported yet, but this makes the getting-started example go through

     $ interop/test.sh
     json file: /var/folders/hz/hgkdl8pj5fg_zlm2qh8x8fy40000gr/T/tmp6iaqtmi8.json
       4 = nki.language.load(%a_input<float64:128,512>)
       5 = nki.language.load(%b_input<float64:128,512>)
       6 = nki.language.add(4,5)
       7 = nki.language.store(%c_output<float64:128,512>,value=6)

Just an idea.


interop/examples/getting_started.py
```diff
@@ -1,15 +1,16 @@
 import nki.language as nl
+import numpy as np
 
 def nki_tensor_add_kernel(a_input, b_input, c_output):
     """NKI kernel to compute element-wise addition of two input tensors
     """
 
     # Check all input/output tensor shapes are the same for element-wise operation
-    assert a_input.shape == b_input.shape == c_output.shape
+    # assert a_input.shape == b_input.shape == c_output.shape
```
govereau, line 9: If you rebase on to my PR, these should work.

Main.lean
```diff
@@ -1,5 +1,33 @@
 import NKL
+import Cli
+import NKL.Python
+open Cli
 
-def main : List String -> IO Unit
-  | [ file ] => IO.FS.readFile file >>= NKL.parse_json
-  | _ => IO.println "invalid arguments"
+def parseJson (p : Parsed) : IO UInt32 := do
```
govereau, line 6: Without the lean_rffi, you don't really need FFI.lean. Could just move NKL.parseJson to Main?

README.md
```diff
@@ -7,3 +7,18 @@ translations from common kernel languages to the KLR core language. The initial
...
+The KLR compiler starts with Python code (e.g. NKI kernels), converts the source
+code to JSON and passes it to the Lean parser. The lean parser converts (aka
+traces) the Python AST into KLR. As such, we have an external dependency on
+a Python runtime. To keep these processes as separate as possible, we just use
+a simple file-IO pipeline;
+
+  1. Python parser parses kernel.py to JSON (using reflection)
+  2. Python writes kernel.json
+  3. KLR reads kernel.json
+  4. KLR writes klr.json
```
govereau, line 23: So, technically you are not writing KLR.json, right?
Do you just need to derive `Lean.ToJson` on the KLR types and call `toJson`?

interop/nkl/parser.py
```diff
@@ -2,16 +2,19 @@
 # Released under Apache 2.0 license as described in the file LICENSE.
 # Authors: Paul Govereau
```
govereau, line 3: Please update the authors when you see them. Should be both of us.

interop/nkl/parser.py
```diff
@@ -118,7 +117,23 @@ def apply_args(self, *args, **kwargs):
...
+    with tempfile.NamedTemporaryFile(mode='w', delete=False, suffix='.json') as temp_file:
+      temp_file.write(json_kernel)
+      temp_file.flush()
+      temp_filename = temp_file.name
+      print("json file: " + temp_filename)
+      klr_filename = temp_filename + ".klr"
+      with open(klr_filename, 'w') as klr_file:
+        dir = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+        try:
+          subprocess.run(dir + "/bin/klr parse-json " + temp_filename, stdout=klr_file, shell=True, check=False)
```
govereau, line 130: Should we give the klr cli an output option, or do you think stdout is sufficient?

govereau (approved):
In general, I think this approach better, it is easier to maintain. I am not entirely convinced that we won't need to integrate more closely with python in the future, but we can cross that bridge if we come to it.

README.md
seanmcl: Yeah, I just copied the text format back to Python for the poc. I'll call `toJson`.

interop/nkl/parser.py
seanmcl: Stdout seems ok for now. Most unix tools don't provide an argument for where to write the output, relying on the shell and other programs like `tee`. We can add it later if this becomes a hassle. Indeed, I was having problems using `"parse-json ... >! foo.json"` as the shell command, so we may want it at some point.

seanmcl: The wheel has a file system, but indeed we'll need to ensure we're pointing to the right spot. I'll try to make a poc wheel today from the lean and python.

seanmcl: We can get it from the wheel like so

```
import subprocess
from importlib.resources import files

def run_binary():
    binary_path = files('your_package').joinpath('../bin/your_binary')
    subprocess.run([str(binary_path), 'arg1', 'arg2'])
```

e.g. when I write

```
# files("numpy")
PosixPath('/Users/seanmcl/.pyenv/versions/3.13.1/envs/klr/lib/python3.13/site-packages/numpy')
```

and 

```
$ ls /Users/seanmcl/.pyenv/versions/3.13.1/envs/klr/lib/python3.13/site-packages/numpy
__config__.py          _array_api_info.py     _pyinstaller           conftest.py            exceptions.py          matlib.py              testing
__config__.pyi         _array_api_info.pyi    _pytesttester.py       core 
...
```

README.md
seanmcl: Now the example test writes

```
$ python3 test/test_examples.py
[{"assign":
  {"x": "4",
   "e":
   {"call":
    {"kwargs": [],
     "f": {"var": {"x": "nki.language.load"}},
     "args":
     [{"tensor":
       {"t":
        {"shape": [128, 512], "name": "a_input", "dtype": "float64"}}}]}}}},
 {"assign":
  {"x": "5",
   "e":
   {"call":
    {"kwargs": [],
     "f": {"var": {"x": "nki.language.load"}},
     "args":
     [{"tensor":
       {"t":
        {"shape": [128, 512], "name": "b_input", "dtype": "float64"}}}]}}}},
 {"assign":
  {"x": "6",
   "e":
   {"call":
    {"kwargs": [],
     "f": {"var": {"x": "nki.language.add"}},
     "args": [{"var": {"x": "4"}}, {"var": {"x": "5"}}]}}}},
 {"assign":
  {"x": "7",
   "e":
   {"call":
    {"kwargs": [["value", {"var": {"x": "6"}}]],
     "f": {"var": {"x": "nki.language.store"}},
     "args":
     [{"tensor":
       {"t":
        {"shape": [128, 512], "name": "c_output", "dtype": "float64"}}}]}}}}]
```

Good enough for now?

govereau: Seems ok. I am working out the BIR output now, it will be similar.

PR #26, opened 2025-02-04
feat: extend builtin object capabilities

govereau:
The main change in this patch is to extend the Object structure with new functions for indexing and binary operators. In addition, the operator names have been converted from strings to enumerations, and the translation of index expressions has been reworked to allow more flexibility. For instance, the new implementation can now support `t[f()]` where f is an arbitrary user function or builtin.

### DIFF
--- a/NKL/Trace/Basic.lean
+++ b/NKL/Trace/Basic.lean
@@ -65,32 +65,33 @@ def indexUnOp : String -> KLR.IndexExpr -> Err KLR.IndexExpr
 
 -- Truthiness of Terms following Python
 
-def Term.isTrue : Term -> TraceM Bool
+def Term.isTrue : Term -> Err Bool
   | .object _ => return true
   | .tuple [] => return false
   | .tuple _  => return true
   | .list []  => return false
   | .list _   => return true
+  | .ellipsis => return true
+  | .slice _ _ _ => return true
   | .expr (.const c) _ => return c.isTrue
-  | .expr _          _ => throw "non-constant expression"
+  | .expr _ _ => throw "non-constant expression"
 
-def Term.isFalse (t : Term) : TraceM Bool :=
+def Term.isFalse (t : Term) : Err Bool :=
   return not (<- t.isTrue)
 
 -- Following Python semantics, boolean operators return
 -- the first value that is convertible to True or False
 
-def boolOp (op : String) (es : List Term) : TraceM Term := do
+def boolOp (op : BoolOp) (es : List Term) : Err Term := do
   bop (<- bopFn op) es
 where
-  bop fn : List Term -> TraceM Term
+  bop fn : List Term -> Err Term
     | []  => throw "invalid expression"
     | [x] => return x
     | x :: xs => do if (<- fn x) then return x else bop fn xs
-  bopFn : String -> TraceM (Term -> TraceM Bool)
-    | "Or"  => return Term.isTrue
-    | "And" => return Term.isFalse
-    | s     => throw s!"unsupported boolean operator {s}"
+  bopFn : BoolOp -> Err (Term -> Err Bool)
+    | .or  => return Term.isTrue
+    | .and => return Term.isFalse
 
 -- Binary Operators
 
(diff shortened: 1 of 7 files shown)


govereau:
Note to reviewers: you may want to start with the changes to `Trace/Types.lean` and `Trace/Python.lean`, everything else is a side-effect of these changes.

NKL/Trace/Types.lean
```diff
@@ -66,43 +66,80 @@ open NKL.KLR
...
-  name : Name
-  type : TermType
-  attr : String -> Err Term
-  call : List Expr -> List (String × Expr) -> Err Term
+  name   : Name
+  type   : TermType := .obj name
+  attr   : String -> Err Term
+  access : List Index -> Err Term
+  index  : TermType -> Nat -> Err Index
+  binop  : Bool -> BinOp -> Expr -> Err Term
```
seanmcl, line 106: What's the bool?

NKL/Trace/Types.lean
```diff
@@ -113,17 +150,25 @@ where
...
-  | .expr _ ty  => return ty
+def Term.type : Term -> TermType
+  | .object obj  => obj.type
+  | .tuple l     => .tuple (types l)
+  | .list l      => .list (types l)
+  | .ellipsis    => .obj "ellipsis".toName
+  | .slice _ _ _ => .obj "slice".toName
+  | .expr _ ty   => ty
+where
+  types : List Term -> List TermType
```
seanmcl, line 161: unused? If we use it we could inline as `xs.map type`

NKL/Trace/Basic.lean
```diff
@@ -103,24 +104,24 @@ where
...
   | .int i      => return List.flatten $ List.replicate i.toNat l
   | _           => throw "invalid multiply"
 
 -- Binary operators on constants
-private def constOp : String -> Const -> Const -> TraceM Term
-  | "Add",  .int l, .int r => return int (l + r)
-  | "Sub",  .int l, .int r => return int (l - r)
-  | "Mult", .int l, .int r => return int (l * r)
-  | "Div",  .int l, .int r => return int (l / r)
+private def constOp : BinOp -> Const -> Const -> Err Term
```
seanmcl, line 114: This works bc both Lean and Python use arbitrary precision integers. Maybe worth mentioning here. I needed to check in Python, since I didn't know off the top of my head.

NKL/Trace/Basic.lean
```diff
@@ -103,24 +104,24 @@ where
 --   [1,2] * True  => [1,2]
 --   [1,2] * False => []
 
-private def mulseq (l : List a) : Const -> TraceM (List a)
+private def mulseq (l : List a) : Const -> Err (List a)
```
seanmcl, line 107: Makes me wonder how much more of these weird Pythonisms we need to handle. Do you think we're close?

NKL/Trace/Python.lean
```diff
@@ -211,8 +221,17 @@ partial def integer (e : Expr) : Tracer Int := do
   | .const c => return (<- c.toInt)
   | _ => throw "expecting integer"
 
+-- Note, this is technically more permissive than python in some contexts (slices)
+partial def integer? : Option Expr -> Tracer (Option Int)
```
seanmcl, line 225: Can we have this just take `Expr` and flatMap it on values of type `Option Expr`? Or does that not work nicely with the stacked monads here?

seanmcl (approved):

NKL/Trace/Types.lean
govereau: The bool is if you are on the left or the right. Open to better ideas.

NKL/Trace/Types.lean
govereau: You lose termination if you use `map`. I could make another gadget like for `mapM`, but seemed easier to just inline map.

NKL/Trace/Basic.lean
govereau: Good call. I will add a comment.

NKL/Trace/Types.lean
seanmcl: I don't understand. What is an example Python thing a binop represents?

NKL/Trace/Basic.lean
govereau: I think we are close. Actually, I haven't seen any kernels using this particular expression, just seemed easy to support.

There are two big things left on my list. Some of the kernels in the `_private_kernels` directory use comprehensions, those are easy to expand out to for loops, but the syntax is a little complicated. Also, some of the kernels use nested functions, but I haven't found any that are actual closures, so we could support this easily. In both cases, the user could just define the thing at the top-level and it will work with no problems (comprehensions would just look like dict literals to us, and we already handle top-level functions). All the cases I have seen could be lifted out and still work---these things tend not to depend on the kernel parameters.

NKL/Trace/Python.lean
govereau: In theory we do not need this `Expr` wrapped in Option at all, because we have `none` as a constant in the target language. So, we could get rid of the outer pattern match on `Option` all together if we do a better job with the Json parsing.

The problem was that sometimes the Json encoding of `None` is `null` and sometimes it is just a missing field, and sometimes it is the literal `None`. I think all these cases could be worked out in the Json parser, but it was easier to allow, `none`, as well as `some (.const .none)`, coming from Python to both be `none`. I have to deal with this in a couple of places, so it should probably be fixed properly.

NKL/Trace/Types.lean
govereau: Note, Python uses multiple functions, e.g. `add` and `radd` (right add), but that didn't seem much better.

govereau: For example:
```
nl.arange(128)[:,None] * i
```
The `*i` part is a `binop` on the object returned by the `access` method of the object returned by the `call` method of the `arange` global.

PR #32, opened 2025-02-06
POC: Create a Python wheel

seanmcl:
This creates a wheel in the pypi test repository.
It is OSX only (or whatever architecture you run it on) so it is not ready for prime time, but can be used for demos and gives the guts of what a real solution will look like.

### DIFF
--- a/interop/klr/parser.py
+++ b/interop/klr/parser.py
@@ -7,8 +7,8 @@ import inspect
 import json
 import numpy as np
 import os
+import pathlib
 import subprocess
-import sys
 import tempfile
 import types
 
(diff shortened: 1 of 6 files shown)


interop/klr/parser.py
```diff
@@ -129,23 +128,29 @@ def apply_args(self, *args, **kwargs):
...
+        temp_file.write(json_kernel)
+        temp_file.flush()
+        temp_filename = temp_file.name
+        # klr_filename needs to be deleted after running klr
+        klr_filename = temp_filename + ".klr"
+        with open(klr_filename, 'w') as klr_file:
+          run_klr(temp_file, klr_file)
+          with open(klr_filename, 'r') as file:
+            klr = file.read()
+            # This isn't great, but when the Lean exe discovers an error, e.g. an undefined
```
govereau, line 146: Right, the old code was catching the IO exception and raising it as a python exception. We need to catch the IO exception and... I guess print it out?

govereau: Or, perhaps this is where we should use stderr?

govereau (approved):

interop/klr/parser.py
seanmcl: I think we will need those errors to be very nicely structured for bubbling up errors to the right spot. So the Lean code would need to do a good job making them a struct with locations and such. For now let's leave it since it seems just doing sth with stderr isn't going to help much. At least all the pytest tests pass now, and we'll make those better with error handling, rather than just ensuring an exception is raised.

PR #33, opened 2025-02-06
Run pytest in CI

seanmcl:
I moved some files around to make the Python structure a little more conventional.

See, e.g., [this build](https://github.com/leanprover/KLR/actions/runs/13188333314/job/36815675807?pr=33)

which shows a successful run of pytest.


govereau (approved):
Nice!

PR #34, opened 2025-02-06
Run CI on OSX and Linux

seanmcl:
WIP to get a build run


govereau (approved):
You are making us look like pros.

PR #35, opened 2025-02-08
fix: report correct line number in error messages

govereau:


interop/klr/parser.py
```diff
@@ -177,29 +179,35 @@ def reference(self, env, refname, val):
     if f is None:
       return
     try:
-      match ast.parse(dedent(inspect.getsource(f))):
+      # get the source lines and starting offset in file
+      src, line = inspect.getsourcelines(f)
+      # remove leading indentation to prevent parse failures
+      # for nested or class functions
+      src = dedent(reduce(add, src))
```
seanmcl, line 186: `"".join(src)` seems a little less highbrow :)

seanmcl (approved):
Neat!

interop/klr/parser.py
govereau: Ha, yes. Switching back and forth from Lean to Python...

PR #36, opened 2025-02-08
feat: infer arguments if not available

govereau:


KLR/Python.lean
```diff
@@ -147,6 +147,11 @@ structure Args where
 def Args.names (args : Args) : List String :=
   args.posonlyargs ++ args.args ++ args.kwonlyargs
 
+-- TODO: does not compute required keyword args
+def Args.required (args : Args) : List String :=
+  let pargs := args.posonlyargs ++ args.args
+  (pargs.reverse.drop args.defaults.length).reverse
```
seanmcl, line 153: `pargs.take (pargs.length - args.defaults.length)`?

KLR/Python.lean
```diff
@@ -184,6 +189,30 @@ structure Kernel where
...
+  if k.args.length > 0 then k else
+    match inferArgs with
+    | none => k
+    | some args => { k with args := args }
+where
+  inferArgs : Option (List Expr') := do
+    let f <- k.funcs.lookup k.entry
+    let args := f.args.required
+    let ten := Expr.exprPos (.const (.int 10)) {}
+    let tensors := args.map fun _ => .tensor [ten,ten] "float32"
```
seanmcl, line 213: Do we need to argue/test/have a tenet somewhere that the shape of the inputs does not impact tracing/compilaton to BIR?

KLR/Python.lean
govereau: Ah, yes, of course.

KLR/Python.lean
govereau: I think we can argue that if the tensors fit in the SBUF, and are valid for the operators, then it doesn't matter exactly what size they are. If tiling is going to happen, then we can get more instructions, but I think the memory regions (a.k.a. intermediate tensors) will be the same in number.

So, we may need to have two modes, one that generates inputs that fit in SBUF, and one that generates inputs that are too big to test out tiling mechanics in the kernel. Some kernels may not handle tiling, so I think we need two modes. Likely this would have to be a command line parameter since it would be hard to determine by looking at the code if it does tiling.

seanmcl: Why 10x10? I guess I'd pick a 0D tensor (e.g. `np.arange(0)` or `np.ndarray(0)` with shape [0] or an array scalar (`np.array(0)` with shape []).

govereau: It seems to me the most common case will be a 2d tensor as most of the ISA instructions operate on 2d tensors. Perhaps 128x128 is a more normal size?

I think if we pick something too trivial, then the kernel will just become a static computation and no ISA instructions will be emitted; or an error will be generated. For instance: `a+1`, which is equivalent to:
```
  nisa.tensor_scalar(a, np.add, 1)
```
is invalid if `a` isn't 2 dimensional.

seanmcl (approved):
I'm ok with this for the demo. I am uncomfortable with trying to guess the shapes of the inputs, but it's a 2-way door.

govereau:
> I'm ok with this for the demo. I am uncomfortable with trying to guess the shapes of the inputs, but it's a 2-way door.

I agree, we need something better. I will merge this for the demo, and we can revisit next week.

PR #37, opened 2025-02-08
CI: Make a wheel

seanmcl:
This took a lot of time to figure out, but we can generate linux and osx wheels now on GitHub Actions that seem functional on both systems. Will probably hit problems, e.g. with different Mac OSs, but I suspect that the right tweaks cibuildwheel arguments will be able to fix those.

[Example run](https://github.com/leanprover/KLR/actions/runs/13218433309) (among a graveyard of failed attempts).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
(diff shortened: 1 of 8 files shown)


govereau (approved):
This is awesome!

PR #43, opened 2025-02-12
feat: track tensor allocations during tracing

govereau:
This patch adds the ability to track needed tensor allocations during the tracing process. To do this, a new `store` statement and term expression are added.  A more complete implementation of the tensor_scalar operator is included as a test case for the new mechanism.

### DIFF
--- a/KLR/Core/Basic.lean
+++ b/KLR/Core/Basic.lean
@@ -3,7 +3,6 @@ Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Paul Govereau, Sean McLaughlin
 -/
-import Lean
 
 /-!
 # Abstract syntax of Core NKL language
@@ -11,31 +10,73 @@ import Lean
 This language is the result of "tracing", and is used as the
 portable format, a.k.a. Kernel Language Representation (KLR).
 -/
-
 namespace KLR.Core
 
--- TODO switch to TensorLib's version of these types
---export TensorLib (Tensor Dtype Shape)
+-- Compute Engines
 
-abbrev Dtype := String
-abbrev Shape := List Int
+inductive Engine where
+  | unassigned
+  | pool
+  | act
+  | pe
+  | dma
+  | dve
+  | sp
+  deriving BEq, Repr
+
+-- ALU operations
+-- TODO organize these into groups
+inductive AluOp where
+  | abs
+  | add
+  | arith_shift_left
+  | arith_shift_right
+  | average
+  | bitwise_and
+  | bitwise_not
+  | bitwise_or
+  | bitwise_xor
+  | bypass
+  | divide
+  | elemwise_mul
+  | is_equal
+  | is_ge
+  | is_gt
+  | is_le
+  | is_lt
+  | logical_and
+  | logical_or
+  | logical_shift_left
+  | logical_shift_right
+  | logical_xor
+  | max
+  | min
+  | mod
+  | mult
+  | not_equal
+  | pow
+  | rsqrt
+  | subtract
+  deriving BEq, Repr
 
 /-
-A TensorName is essentially a typed variable, where the type
-must be a tensor type. When we flush out Typ below we may replace
-this with `Expr.var name (Typ.tensor dtype shape)`. For now, this
-only refers to dynamic tensors, or compile-time tensors, not
-trace-time tensors.
+A TensorName is essentially a typed variable, where the type must be a
+tensor type. This only refers to dynamic tensors, or compile-time
+tensors, not trace-time tensors.
 -/
+abbrev Dtype := String
+abbrev Shape := List Nat
+
+inductive Memory where
+  | dram | sbuf | pmem | reg
+  deriving Repr, BEq
 
 structure TensorName where
   name  : String
   dtype : Dtype
   shape : Shape
-  deriving Repr, BEq, Lean.ToJson
-
--- TODO
-inductive Typ where
+  memory: Memory
+  deriving Repr, BEq
 
 inductive Const where
   | none
@@ -43,7 +84,7 @@ inductive Const where
   | int (value : Int)
   | float (value : Float)
   | string (value : String)
-  deriving Repr, BEq, Lean.ToJson
+  deriving Repr, BEq
 
 -- This corresponds to the "Quasi-Affine Expressions" in Neuron.
 -- Note, `floor` is the usual integer division.
@@ -56,27 +97,72 @@ inductive IndexExpr where
   | floor (expr : IndexExpr) (scalar : Int)
   | ceil (expr : IndexExpr) (scalar : Int)
   | mod (expr : IndexExpr) (scalar : Int)
-  deriving Repr, BEq, Lean.ToJson
+  deriving Repr, BEq
 
 -- Note: `np.newindex` is represented as `(.coord none)`
 inductive Index where
   | ellipsis
   | coord (e : Option IndexExpr)
   | slice (l u step : Option IndexExpr)
-  deriving Repr, BEq, Lean.ToJson
+  deriving Repr, BEq
+
+structure TensorScalar where
+  op0 : AluOp
+  const0 : Float
+  reverse0 : Bool
+  op1 : AluOp
+  const1 : Float
+  reverse1 : Bool
+  deriving Repr, BEq
+
+inductive Operator where
+  | tensorScalar : TensorScalar -> Operator
+  deriving Repr, BEq
 
 inductive Expr where
   | var (x : String)
   | const (c : Const)
   | tensor (t : TensorName)
   | access (t : Expr) (ix : List Index)
+  | operator (op : Operator)
   | call (f : Expr) (args : List Expr) (kwargs : List (String × Expr))
-  deriving Repr, BEq, Lean.ToJson
+  deriving Repr, BEq
 
 inductive Stmt where
-  | pass
-  | expr (v : Expr)
   | ret (v : Expr)
+  | store (t : TensorName) (ix : List Index) (e : Expr)
   | assign (x : String) (e : Expr)
   | loop (x : String) (l u step : IndexExpr) (body : List Stmt)
-  deriving Repr, BEq, Lean.ToJson
+  deriving Repr, BEq
+
+structure Kernel where
+  name : String
+  inputs : List TensorName
+  outputs : List TensorName
+  body : List Stmt
+  deriving Repr, BEq
+
+-- TODO: not efficient
+partial def Expr.tensors : Expr -> List TensorName :=
+  tensors []
+where
+  tensors (l : List TensorName) : Expr -> List TensorName
+  | .var _ => l
+  | .const _ => l
+  | .tensor t => l.insert t
+  | .access t _ => tensors l t
+  | .operator _ => l
+  | .call f args kwargs =>
+      let l := tensors l f
+      let l := args.foldl tensors l
+      kwargs.foldl (fun l kw => tensors l kw.snd) l
+
+partial def Stmt.tensors : Stmt -> List TensorName
+  | .ret e => e.tensors
+  | .store t _ e => e.tensors.insert t
+  | .assign _ e => e.tensors
+  | .loop _ _ _ _ body => (body.map tensors).flatten.eraseDups
+
+def Kernel.internal (k : Kernel) : List TensorName :=
+  let ts := (k.body.map Stmt.tensors).flatten.eraseDups
+  (ts.removeAll k.inputs).removeAll k.outputs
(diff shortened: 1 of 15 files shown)


KLR/Core/Basic.lean
```diff
@@ -3,47 +3,88 @@ Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
...
 namespace KLR.Core
 
--- TODO switch to TensorLib's version of these types
---export TensorLib (Tensor Dtype Shape)
+-- Compute Engines
 
-abbrev Dtype := String
-abbrev Shape := List Int
+inductive Engine where
+  | unassigned
```
seanmcl, line 18: `Option`? I wonder if we'll be doing a lot of `if engine == unassigned then ...do some inference`

KLR/Core/Basic.lean
```diff
@@ -3,47 +3,88 @@ Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
...
-trace-time tensors.
+A TensorName is essentially a typed variable, where the type must be a
+tensor type. This only refers to dynamic tensors, or compile-time
+tensors, not trace-time tensors.
 -/
+abbrev Dtype := String
+abbrev Shape := List Nat
+
+inductive Memory where
+  | dram | sbuf | pmem
```
seanmcl, line 71: There are also registers iiuc. Do we not need those?

KLR/Core/Basic.lean
```diff
@@ -3,47 +3,88 @@ Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
...
   shape : Shape
-  deriving Repr, BEq, Lean.ToJson
-
--- TODO
-inductive Typ where
+  memory: Memory
+  deriving Repr, BEq
 
 inductive Const where
   | none
```
seanmcl, line 82: what is `none` again?

KLR/Core/Basic.lean
```diff
@@ -56,27 +97,72 @@ inductive IndexExpr where
...
+  | .const _ => l
+  | .tensor t => l.insert t
+  | .access t _ => tensors l t
+  | .operator _ => l
+  | .call f args kwargs =>
+      let l := tensors l f
+      let l := args.foldl tensors l
+      kwargs.foldl (fun l kw => tensors l kw.snd) l
+
+partial def Stmt.tensors : Stmt -> List TensorName
```
seanmcl, line 160: Maybe make it a `Set` instead of all the dup detection?

KLR/Core/FromToJson.lean
```diff
@@ -0,0 +1,88 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Paul Govereau, Sean McLaughlin
+-/
+import Lean
+import KLR.Core.Basic
+
+/-
+# Instances of ToJson for KLR
+
+The instances are placed here, in a separate module, because we need to
```
seanmcl, line 12: Nice comment. What does 'interfere with other code' mean though?

Main.lean
```diff
@@ -1,24 +1,35 @@
...
+  let klr <- KLR.Trace.runNKIKernel kernel.inferArguments
+  if p.hasFlag "repr" then
+    IO.println (toString $ repr klr)
+  else if p.hasFlag "json" then
+    IO.println (toString $ Lean.toJson klr)
+  else
+    IO.println (Lean.format klr).pretty
+  return 0
+
 def parseJsonCmd := `[Cli|
```
seanmcl, line 33: Let's mark this as deprecated, since it seems subsumed by `trace`. Will need to push the change through the interop code but that should be a one liner.

KLR/Trace/Types.lean
```diff
@@ -113,14 +113,32 @@ to a set of tensor indexes, or to a pass statement depending on context.
...
-TODO: KLR expressions should have KLR types, not TermTypes
+
+The store expression is an interesting case. In the original python
+code, a store can show up in an expression context. For example:
+
+  c = nl.load(a) + nl.load(b)
+
+Which is technically:
+
+  c = (store(t1, nl.load(a)) + store(t2, nl.load(b))
```
seanmcl, line 124: Where are `t1, t2` coming from in this example?

KLR/Trace/Tensor.lean
```diff
@@ -13,6 +14,206 @@ TODO: These are just place holders...
...
+  | _ => throw "expecting tensor"
+
+-- This only handles the simple cases
+-- Note: Maybe only simple cases are possible at this point ??
+def inferShape (t : TensorName) : List Index -> Err Shape
+  | [] | [.ellipsis] => return t.shape
+  | ix => do
+      let base := t.shape
+      if base.length != ix.length then
+        throw "unsupported index"
```
seanmcl, line 41: This seems pretty easy to handle by making the missing dimensions "full", but fine for now.

KLR/Trace/Tensor.lean
```diff
@@ -13,6 +14,206 @@ TODO: These are just place holders...
...
+      | (t, [.ellipsis]) => return (t, ix)
+      | _ => throw "unsupported tensor expression"
+  | _ => throw "expecting tensor"
+
+def Term.inferTensor : Term -> Err (TensorName × List Index)
+  | .expr e (.tensor _ _) => Expr.inferTensor e
+  | _ => throw "expecting tensor"
+
+-- This only handles the simple cases
+-- Note: Maybe only simple cases are possible at this point ??
```
seanmcl, line 35: Relies on some assumptions, like `t` can't have 0 dimensions, `x <= i` on the map, etc. We can probably do better here but for demo seems ok.

KLR/Trace/Tensor.lean
```diff
@@ -13,6 +14,206 @@ TODO: These are just place holders...
...
+      let base := t.shape
+      if base.length != ix.length then
+        throw "unsupported index"
+      let dims <- (base.zip ix).mapM fun (x, i) =>
+        match i with
+        | .coord _ => return 0
+        | .slice none none none
+        | .slice (some (.int 0)) none none => return x
+        | .slice none (some (.int i)) none
+        | .slice (some (.int 0)) (some (.int i)) none => return i.toNat
```
seanmcl, line 48: Examples only index starting at 0?

KLR/Trace/Tensor.lean
```diff
@@ -13,6 +14,206 @@ TODO: These are just place holders...
...
+        | _ => throw "unsupported index"
+      return dims.filter (. != 0)
+
+def Expr.inferShape (e : Expr) : Err Shape := do
+  let (t, ix) <- inferTensor e
+  Tensor.inferShape t ix
+
+-- declare a new tensor, unique to the current expression
```
seanmcl, line 56: I don't understand why we need this. Example?

KLR/Trace/Tensor.lean
```diff
@@ -13,6 +14,206 @@ TODO: These are just place holders...
...
+def Int.fromNKI : Term -> Err Int
+ | .expr (.const (.int i)) .int => return i
+ | _ => throw "expecting integer"
+
+def Nat.fromNKI (t : Term) : Err Nat :=
+  return (<- Int.fromNKI t).toNat
+
+def Float.fromNKI : Term -> Err Float
+ | .expr (.const (.float f)) .float => return f
+ | _ => throw "expecting integer"
```
seanmcl, line 114: expecting float

KLR/Trace/Tensor.lean
```diff
@@ -13,6 +14,206 @@ TODO: These are just place holders...
...
+
+def Memory.fromNKI (t : Term) : Err Memory :=
+  match t.type with
+  | .obj name =>
+    match name.toString with
+    | "nki.language.shared_hbm" => return .dram
+    | "nki.language.sbuf" => return .sbuf
+    | "nki.language.pmem" => return .pmem
+    | _ => throw "expecting buffer type"
+  | t => throw s!"expecting buffer type 2 got {repr t}"
```
seanmcl, line 128: "buffer type 2"?

KLR/Trace/Tensor.lean
```diff
@@ -13,6 +14,206 @@ TODO: These are just place holders...
...
+def ndarray : GlobalFn :=
+  withArgs [("shape", none),
+            ("dtype", none),
+            ("buffer", some_string "nki.language.sbuf")]
+  fun
+  | [shape, dtype, buf] => do
+      let shape <- Shape.fromNKI shape
+      let dtype <- String.fromNKI dtype
+      let memory <- Memory.fromNKI buf
+      let t <- declare "t" dtype shape memory
```
seanmcl, line 159: "t"?

KLR/Trace/Builtin.lean
```diff
@@ -14,12 +14,40 @@ import KLR.Trace.Types
...
+  let extra := names.drop pos.length
+  let kws <- extra.mapM fun (name, dflt) => do
+    match kw.find? (fun p => p.fst == name) with
+    | some x => return x.snd
+    | none => match dflt with
+              | some x => return x
+              | none => throw s!"argument {name} not found"
+  return pos.append kws
+
+def withArgs [Monad m] [MonadExcept String m] [MonadLift Err m]
```
seanmcl, line 46: This function does some heavy lifting. Nice. Took a while for me to appreciate it.

KLR/Trace/Tensor.lean
```diff
@@ -13,6 +14,206 @@ TODO: These are just place holders...
...
+  withArgs [("data", none),
+            ("op0", none),
+            ("operand0",none),
+            ("reverse0", some_bool false),
+            ("op1", some_none),
+            ("operand1", some_none),
+            ("reverse1", some_bool false),
+            ("dtype", some_none)]
+  fun
+  | [data,op0,operand0,reverse0,op1,operand1,reverse1,dtype] => do
```
seanmcl, line 195: nit: spaces here please. Hard to read.

seanmcl (approved):
Neat stuff. Looking fancier.

KLR/Core/Basic.lean
govereau: This is a BIR type: the compiler recognizes "unassigned" as a compute engine, and then does an automatic assignment. It is possible we will never use this feature, then we could delete it.

KLR/Core/Basic.lean
govereau: There are registers, and they have a memory type, I will add it.

KLR/Core/Basic.lean
govereau: This is Python `None`

KLR/Core/FromToJson.lean
govereau: I was worried about overriding the Float instances. So, unless you import this file you don't get new instances for Float.

KLR/Trace/Types.lean
govereau: These are those invented names we were discussing, `load.4.12`, etc. Basically each load expression gets a tensor allocated to it.

KLR/Trace/Tensor.lean
govereau: Yeah, I keep running into little things that I am almost sure you have already implemented in TensorLib.

KLR/Trace/Tensor.lean
govereau: Yup, in general, this is much more complex. I am still not convinced we have all the design around indexing figured out.

KLR/Trace/Tensor.lean
govereau: Very lazy. I don't have any examples that are not zero. The test suite for NKI isn't very comprehensive in this way.

KLR/Trace/Tensor.lean
govereau: thx.

KLR/Trace/Tensor.lean
govereau: Adding this comment to code:

We have to calculate how much memory we are going to use. The most memory we
could need is one tensor for each expression in the source program (loops can
reuse the same memory in each iteration). This code gives the name of the
tensor associated with an expression. Right now I am using the source location,
but we could also use a hash of expression, or something else. Note, this
expression may be evaluated many times, and we want this function to always
return the same result.

In the end we walk over the KLR kernel and collect all the TensorNames, and
these represent the memory we need to allocate in the dram, sbuf, etc.

KLR/Trace/Tensor.lean
govereau: Ah, should be "ndarray".

KLR/Trace/Tensor.lean
govereau: Oops, that was debugging. removed.

KLR/Trace/Builtin.lean
govereau: I should put a comment saying what it is for, at least.

KLR/Core/Basic.lean
govereau: I couldn't find set. I think it is in Batteries?

PR #44, opened 2025-02-12
ci: Add arm build host

seanmcl:
We currently aren't building ARM wheels. I'm not sure this will work, but let's try it. We only push the wheels on the main branch, so we need to merge this to try it.

Build works:
https://github.com/leanprover/KLR/actions/runs/13293628967/job/37120447588


seanmcl:
The mac fleet seems overutilized; some jobs are just hanging, but all the linux ones are going through.

govereau (approved):

PR #47, opened 2025-02-12
feat: add BIR definitions

govereau:
Add definitions for BIR and skeleton of translation from KLR.


Main.lean
```diff
@@ -49,13 +72,35 @@ def traceCmd := `[Cli|
...
+  "compile" VIA compile;
+  "Compile Python to BIR"
+
+  FLAGS:
+    r, repr; "Output Repr format"
+  ARGS:
+    file : String; "File of Python AST printed as JSON"
+]
+
 def klrCmd : Cmd := `[Cli|
```
seanmcl, line 95: Little CLI is looking nice!

Nit: can we sort the command names?

KLR/BIR/Instruction.lean
```diff
@@ -0,0 +1,122 @@
...
+export KLR.Core (Dtype Shape Engine AluOp Memory)
+
+-- TODO: this is incomplete
+structure QAETerm where
+  coeff: Int
+  deriving BEq, Repr, Lean.FromJson, Lean.ToJson
+
+structure QuasiAffineExpr where
+  kind : String := "AffExprKind"
+  --terms: List QAETerm := []
```
seanmcl, line 28: are commented field just not used but we want to uncomment eventually?

KLR/BIR/Instruction.lean
```diff
@@ -0,0 +1,122 @@
...
+  --c : Int := 0
+  deriving BEq, Repr, Lean.FromJson, Lean.ToJson
+
+abbrev ShapeVector := List Nat
+abbrev LoopAxis := String
+
+inductive InstSyncType where
```
seanmcl, line 35: Do the inductive names matter, or only the enum names?

KLR/BIR/Instructions.lean
```diff
@@ -0,0 +1,940 @@
+/-
+Copyright (c) 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
seanmcl, line 2: Which part of this file is generated by brewer?

KLR/BIR/Function.lean
```diff
@@ -0,0 +1,199 @@
...
+instance : Lean.ToJson Block where toJson := Block.toJson
+instance : Lean.ToJson Loop where toJson := Loop.toJson
+
+mutual
+
+partial def InstLoop.fromJson? (j : Lean.Json) : Err InstLoop := do
+  let opcode <- j.getObjValAs? String "opcode"
+  match opcode with
+  | "Loop" => return .L (<- Loop.fromJson? j)
+  | _ => return .I (<- Lean.fromJson? j)
```
seanmcl, line 157: Could we have a comment what is going on here? Why is this the only opcode that needs to be patched?

seanmcl (approved):

Main.lean
govereau: Yup, sure thing.

KLR/BIR/Instruction.lean
govereau: Not used, but will be needed. In this case, I don't think we will need the full power of QAEs, probably we just need constants. Remains to be seen.

KLR/BIR/Instruction.lean
govereau: They do if we are using the derived To- and FromJson instances.

KLR/BIR/Instructions.lean
govereau: The entire Instructions.lean file is generated. The others are hand-written.

KLR/BIR/Function.lean
govereau: I will add a comment.

PR #48, opened 2025-02-13
refactor: organize the BIR compiler files

govereau:


KLR/BIR/Compile/Memory.lean
```diff
@@ -0,0 +1,99 @@
...
+/-
+# Memory allocation
+
+Create Allocations for each named tensor. The access pattern compiler (below)
+needs to know how the physical memory is defined for each tensor. For now, all
+tensors have the same basic layout.
+-/
+
+-- TODO: just a for instance...
+def physicalShape (t : TensorName) : Shape :=
```
seanmcl, line 24: This shouldn't be called `shape` I think. Maybe some new word like `size` or `layout` or something.

KLR/BIR/Compile/Memory.lean
```diff
@@ -0,0 +1,99 @@
...
+    | .Input, _ => .Input
+    | .Output, _ => .Output
+    | _, .dram => .DRAM
+    | _, .sbuf => .SB
+    | _, .pmem => .PSUM
+    | _, .reg => .REG
+  return {
+    addr_space := some .Shared  -- LNC ?
+    dtype := some t.dtype
+    partition_dim := some 0
```
seanmcl, line 43: is it always 0? Do we need to track this through TensorName?

KLR/BIR/Compile/Memory.lean
```diff
@@ -0,0 +1,99 @@
...
+
+Generate access patterns for tensor access expressions.
+-/
+
+-- Just to get things working, here are variations for full 2d tensors only
+-- e.g. t[:,:] or t[...] or similar
+def dimToAP (d1 d2 : Nat) : Compile PhysicalAccessPattern :=
+  return {
+    ap := [ [ d2, d1 ], [ 1, d2 ] ]
+    dtype := "" -- filled in below
```
seanmcl, line 65: Do we want to do better on the types here? Or are strings good?

KLR/BIR/Compile/Memory.lean
```diff
@@ -0,0 +1,99 @@
...
+    offset := 0
+    memsetref := "" -- filled in below
+    memref := "" -- filled in below
+  }
+
+def slicesToAP (d1 d2 : Nat) : List Index -> Compile PhysicalAccessPattern
+  | [ .ellipsis ] => dimToAP d1 d2
```
seanmcl, line 72: nit, duplicate return value

KLR/BIR/Compile/Memory.lean
```diff
@@ -0,0 +1,99 @@
...
+  | [ .slice none none none, .slice none none none ] => dimToAP d1 d2
+  | [ .slice none (some (.int a)) none,
+      .slice none (some (.int b)) none ] => do
+        if a < 0 || b < 0 then
+          throw "negative slice patterns not supported"
+        let a := a.toNat
+        let b := b.toNat
+        if a < d1 || b < d2 then
+          throw "partial slice patterns not supported"
+        dimToAP a b
```
seanmcl, line 82: I don't understand this. This should just be another case ofo `dimToAP d1 d2` right?

KLR/BIR/Compile/Memory.lean
```diff
@@ -0,0 +1,99 @@
...
+  | _ => throw "unimplemented access pattern"
+
+private def setMemRef (t : TensorName) (ap : PhysicalAccessPattern) : PhysicalAccessPattern :=
+  { ap with
+    dtype := t.dtype
+    memsetref := t.name ++ "_set"
```
seanmcl, line 88: where is this convention coming from? Is it a BIR thing?

KLR/BIR/Instruction.lean
```diff
@@ -77,12 +77,29 @@ structure ImmValue where
   value : Lean.JsonNumber
   deriving BEq, Repr, Lean.FromJson, Lean.ToJson
 
+structure ImmArray where
```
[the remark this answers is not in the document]
govereau: Immediate array, in the sense of a constant such as `1.0`. (BIR name)

KLR/BIR/Instruction.lean
```diff
@@ -77,12 +77,29 @@ structure ImmValue where
...
+  dtype : String
+  value_array : List Lean.JsonNumber
```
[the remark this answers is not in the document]
govereau: Yes, we do. There are 5 types allowed: f32, i32, u8, u32, u64. I don't understand how the architecture effects this, so I left it general for now.

KLR/BIR/Compile/Memory.lean
govereau: Agreed. BIR conflates the types, but we should fix.

KLR/BIR/Compile/Memory.lean
govereau: Yes, it can be different, but we do not support this in the tracer, yet.

KLR/BIR/Compile/Memory.lean
govereau: Yes, I think we want `TensorLib.Dtype` ?
But it should match what the HW has, so maybe we should think about it a bit.

KLR/BIR/Compile/Memory.lean
seanmcl: Can we have a comment then?

KLR/BIR/Compile/Memory.lean
govereau: Nice, changed.

KLR/BIR/Compile/Memory.lean
govereau: Hmm, I don't know. (I was going to ask you)

KLR/BIR/Compile/Memory.lean
seanmcl: If you agree with comments, but don't want to change them now, could we just add a comment, e.g. `-- TODO: hange the name. This is not really a shape in the sense it's used in array programming.` or something like that.

KLR/BIR/Compile/Memory.lean
govereau: Penguin convention.

seanmcl (approved):

KLR/BIR/Compile/Memory.lean
seanmcl: Yes, if you over-index the array it's ok, but will be the same as if you but the size.

```
In [37]: x = np.arange(10)

In [38]: x[0:11:1]
Out[38]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
```

KLR/BIR/Compile/Memory.lean
seanmcl: could we move it to a util function then? It happens twice, and is a cryptic magic constant.

PR #49, opened 2025-02-13
test: add some test kernels for tensor_scalar

govereau:


KLR/Trace/Tensor.lean
```diff
@@ -188,7 +192,7 @@ def store : GlobalFn :=
   fun
   | [.expr dst (.tensor _ s₁), .expr src (.tensor _ s₂)] => do
       if s₁ != s₂ then
-        throw "incompatible shapes {s₁} {s₂}"
+        throw s!"incompatible shapes {s₁} {s₂}"
```
seanmcl, line 195: Urg. Lean should have a linter warning for this.

interop/test/test_nki_isa_tensor_scalar.py
```diff
@@ -0,0 +1,60 @@
...
+Unit tests for tensor_scalar.
+
+If using these tests for development, you can generate the NKI.json from
+the top-level like so:
+
+  PYTHONPATH=interop:interop/test ./bin/gather test_nki_isa_tensor_scalar.kernel1 > kernel1.json
+
+and then, e.g.
+
+  lake exe klr trace kernel1.json
```
seanmcl, line 20: Optional: or `./bin/klr`, which avoids overhead of rechecking the build each time.

```
$ time bin/klr trace /tmp/kernel1.json >/dev/null
0.03s user 0.01s system 51% cpu 0.073 total
$ time lake exe klr trace /tmp/kernel1.json >/dev/null
0.15s user 0.14s system 65% cpu 0.439 total
```

interop/test/test_nki_isa_tensor_scalar.py
```diff
@@ -0,0 +1,60 @@
...
+  lake exe klr compile kernel1.json
+"""
+
+# utility function - allocate memory in DRAM
```
seanmcl, line 24: Just a curiousity: Why does everyone say DRAM when it's going on HBM, and there's also DRAM attached to the main OS. I'd think HBM would be more clear. I guess it's always unambiguous which DRAM is which?

interop/test/test_nki_isa_tensor_scalar.py
```diff
@@ -0,0 +1,60 @@
...
+def alloc_like(t):
+  return nl.ndarray(t.shape, dtype=t.dtype, buffer=nl.shared_hbm)
+
+# utility function - allocate memory in DRAM and copy SBUF tile to it
+def dram_tile(a):
```
seanmcl, line 29: should we check `a` is on sbuf?

interop/test/test_nki_isa_tensor_scalar.py
```diff
@@ -0,0 +1,59 @@
...
+  b = alloc_like(a)
+  nl.store(b, a)
+  return b
+
+# test kernel 1 : t - 1.0 with no access pattern
+def kernel1(a):
+  a_tile = nl.load(a)
+  b_tile = nisa.tensor_scalar(a_tile, np.subtract, 1.0)
```
seanmcl, line 37: Should our linter give an error on this? `np.subtract` [doesn't appear supported](https://awsdocs-neuron.readthedocs-hosted.com/en/latest/general/nki/api/nki.api.shared.html#nki-aluop).

seanmcl (approved):

interop/test/test_nki_isa_tensor_scalar.py
govereau: Umm, I think that is just me using the wrong word. I guess I am a little confused in my mind between the DRAM on the host and the HBM on the accelerator.

interop/test/test_nki_isa_tensor_scalar.py
govereau: Actually, I asked about this in a slack thread. It is supported, and most of the examples use `numpy` ufuncs in these locations, but I proposed we create an enum in python that is more explicit.

interop/test/test_nki_isa_tensor_scalar.py
govereau: Yeah, that would be good.

interop/test/test_nki_isa_tensor_scalar.py
govereau: Oh yeah, good call.

interop/test/test_nki_isa_tensor_scalar.py
govereau: Actually, it breaks the tracer, so I will add it along with a tracer fix (it is minor).

interop/test/test_nki_isa_tensor_scalar.py
seanmcl: to be fair, it executing in under 3 seconds is a joy coming from mostly Java/Scala programming.

PR #50, opened 2025-02-13
feat: Detect undefined symbols

seanmcl:
Our current Python parser does not notice undefined symbols. While we don't do anything with them now, we want to pass that info back to the caller in the forthcoming linter. In addition, we intended parsing to fail when there are undefined symbols, as indicated in test_basic.py, but it was failing for an unrelated reason before, and broke a few commits ago.

This change just makes a best effort attempt to notice undefined symbols, and crashes accordingly in the klr binary. In the future we will do more with this.

This change fixes the broken CI build.

### DIFF
--- a/KLR/Python.lean
+++ b/KLR/Python.lean
@@ -183,12 +183,16 @@ An example of a global is:
     else:
       ...
 -/
+instance : Repr Lean.Json where
+  reprPrec jsn _ := jsn.compress
+
 structure Kernel where
   entry : String
   funcs : List (String × Fun)
   args : List Expr'
   kwargs : List (String × Expr')
   globals : List (String × Expr')
+  undefinedSymbols : List Lean.Json
   deriving Repr
 
 /-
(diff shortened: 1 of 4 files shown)


KLR/Python.lean
```diff
@@ -183,12 +183,16 @@ An example of a global is:
     else:
       ...
 -/
+instance : Repr Lean.Json where
+  reprPrec jsn _ := jsn.compress
```
govereau, line 187: Nice.

interop/klr/parser.py
```diff
@@ -250,3 +260,13 @@ def visit_Attribute(self, node):
       raise e
     except Exception:
       return
+
+
+if __name__ == '__main__':
```
govereau, line 265: I was thinking we might do: run_klr here?

govereau: And then add a gather subcommand?

govereau (approved):

interop/klr/parser.py
govereau: Actually, not here. in wherever `python -m klr-lang` goes to.

seanmcl: ooh good idea. Another PR.

PR #51, opened 2025-02-14
feat: tracing for basic control-flow

govereau:


KLR/Trace/Python.lean
```diff
@@ -273,35 +305,51 @@ partial def expr' : Expr' -> Tracer Item
...
+      let body := if <- t.isTrue then thn else els
+      stmts body
+  | .forLoop x iter body orelse => do
+      let x <- LValue x
+      let iter <- term iter
+      let ts <- termToIter iter
+      for t in ts do
+        assignTerm x t
+        let res <- stmts body
+        if res == .cont then continue
```
seanmcl, line 340: oof, not sure how we'll prove termination here

KLR/Trace/Python.lean
```diff
@@ -273,35 +305,51 @@ partial def expr' : Expr' -> Tracer Item
 partial def keyword (f : Expr -> Tracer a) : Keyword -> Tracer (String × a)
   | .keyword id e p => withPos p do return (id, (<- f e))
 
-partial def stmt : Stmt -> Tracer (Option Term)
+partial def stmt : Stmt -> Tracer StmtResult
   | .stmtPos s' p => withPos p (stmt' s')
 
-partial def stmt' : Stmt' -> Tracer (Option Term)
-  | .pass => return none
+partial def stmt' : Stmt' -> Tracer StmtResult
```
seanmcl, line 311: why is this partial? Can we try to not use partial definitions? Maybe use `| _ => .error xxx` instead?

seanmcl (commented):
Starting to look like a shallow embedding!

KLR/Trace/Python.lean
```diff
@@ -211,10 +211,42 @@ def assign (xs : List Expr) (e : Term) : Tracer Unit := do
   for x in xs do
     assignTerm x e
 
+-- Translation of for-loop iterators
+
+-- range, but only in a for-loop context
+-- step is not zero (see below)
+partial def range (start stop step : Int) : List Term :=
```
seanmcl, line 218: Ug, I lost a long comment on this. Not crazy about the partial definition. As written termination is tricky bc the natural metric, `|stop - start|` doesn't decrease if the `step` is massive. We know it stops on the next round, but convincing Lean of that is tricky. I tried some fancier metrics with `min` but it made the code complicated. This version works, but is a bit less elegant than yours.

As an aside, the `where` definitions are usually OK, but when there are proof obligations for termination you need to put those before the `where`, right under the recursive definition. That means there is arbitrary tactic script between the code you're reading, and the auxiliary definitions you need. I prefer `let` definitions for this reason.

```
private def range1 (start stop step : Int) : List Term :=
  let int i := Term.expr (.const (.int i)) .int
  if step = 0 then []
  else if 0 < step then
    if stop <= start then [] else
    if stop <= start + step then [int start] else
    int start :: range1 (start + step) stop step
  else -- step < 0
    if start <= stop then [] else
    if start + step <= stop then [int start] else
    int start :: range1 (start + step) stop step
termination_by (stop - start).natAbs
```

KLR/Trace/Python.lean
govereau: It is partial because it is in a mutual block that really is partial.
```
function_call -> stmt -> expr -> function_call
```
termination depends on if the original python code is terminating.
I wish this mutual block was smaller, but I couldn't see how.

Long term, we probably want to detect deeply recursive computations and halt them with a error, then we can probably prove this code terminates within a certain number of steps.

KLR/Trace/Python.lean
govereau: beautiful. I am copying your version in.

KLR/Trace/Python.lean
seanmcl: OK, yeah let's do that. Python has that too right? Either there is a natural termination order, or there is unbounded recursion in which case it would stack overflow since there is no recursion elimination optimizations. That's my gut feeling at least but maybe I'm missing something...

seanmcl: I guess it could do a loop over the arbitrary precision numbers which could be larger than whatever bound we chose. But maybe too niche to care about in practice?

seanmcl (approved):

KLR/Trace/Python.lean
govereau: It seems to me that recursion may be an anti-feature for NKI. There are still loops, but we should be able to make something work.

PR #55, opened 2025-02-14
fix: define range in the environment

govereau:


KLR/BIR/Function.lean
```diff
@@ -78,7 +78,7 @@ structure MemoryLocation where
   base : Nat := 0   -- base partition (0, 32, 64)
   allocated : Option Bool := some false
   pinned : Bool := false
-  tensor_id : Option Nat := none
+  tensor_id : Option Nat := some 0
```
seanmcl, line 81: why optional if it crashes the verifier?

seanmcl (approved):

KLR/BIR/Function.lean
govereau: Well, I found BIR files in the test suite where it was missing. Maybe those are deprecated?

PR #57, opened 2025-02-15
feat: Scripts to generate and run BIR

seanmcl:
Adds some scripts and cli improvement to run gather and nki-to-bir without requiring explicit PYTHONPATH munging.

The run-bir script has some hard-coded paths at the moment, but
we should be able to get more general, or at least give a
list of instructions for getting the right setup as a followup.

This adds `gather` to the wheel, so we can run from the wheel.
For example

    $ lake build
    $ bin/make-wheel
    $ cd /tmp
    $ virtualenv env
    $ source env/bin/activate
    $ pip install ~/workplace/sherloc/KLR/.wheel/dist/klr_lang-0.0.7-cp313-cp313-macosx_14_0_arm64.whl
    $ env/lib/python3.13/site-packages/bin/klr nki-to-bir ~/workplace/sherloc/KLR/interop/test/test_nki_isa_tensor_scalar.py kernel1
    {"version": 2,
     "functions":
     [{"name": "sg0000",
     ...

### DIFF
--- a/bin/run-bir
+++ b/bin/run-bir
@@ -1,0 +1,88 @@
+#!/bin/bash
+set -e -u -o pipefail
+trap "kill 0" SIGINT SIGTERM
+
+# Run a BIR JSON file on a trn EC2 instance using the public Neuron SDK
+# You can create one, for example, using make-bir in this directory.
+# To use this, copy your BIR.json file to the instance and run this
+# script. You will need several tools from the Neuron SDK in your path.
+# A bunch of files get generated, so it's best to do this in a new
+# directory, where the only file is the BIR. E.g. a workable pattern is
+# something like this, where 'trn' is a profile of a trainium instance
+# in your .ssh/config
+#
+#    osx$ scp run-bir trn:bin/run-bir # assumes bin is a dir at user root on the path
+#    osx$ scp bir.json trn:
+#    osx$ ssh trn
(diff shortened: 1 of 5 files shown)


bin/dump-bir
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
seanmcl, line 1: This may be done better in Main.lean, but the way it's written it's not ideal. Maybe we need those functions to be split up a bit and return String so we can combine e.g. gather and compile more easily.

seanmcl: Updated to move it to Lean rather than Python

bin/run-bir
```diff
@@ -0,0 +1,88 @@
...
+
+# TODO: Absolute path to NCC is sketchy here. Find a better way.
+NEURONXCC_DIR=/opt/aws_neuronx_venv_pytorch_2_5/lib64/python3.9/site-packages/neuronxcc
+#TODO: This gets generated during compilation. Still figuring this out.
+TENSOR_MAP_FILE=tensor_map.json
+
+WALRUS_DRIVER=$NEURONXCC_DIR/starfish/bin/walrus_driver
+echo "Logging to $LOG_FILE"
+# Args are from a regular run of NKI baremetal logs
+$WALRUS_DRIVER \
```
govereau, line 37: wow.

govereau (approved):
This is great!

PR #60, opened 2025-02-17
feat: remove dependency on NKI definitions

govereau:
This change makes it possible to trace and compile kernels without needing a copy of the NKI API signatures in Python. This builds on the previous commit that accurately tracks undefined names in the parser.


KLR/Python.lean
```diff
@@ -471,12 +471,11 @@ def kernel (j : Json) : Parser Kernel := do
...
-  let undefinedSymbols <- field (list json) j "undefined_symbols"
+  let undefinedSymbols <- field (list str) j "undefined_symbols"
   return Kernel.mk name funcs args kwargs globals undefinedSymbols
 
 def parse (s : String) : Err Kernel := do
   let jsn <- Json.parse s
   match kernel jsn {} with
-  | .ok x _ =>
-    if !x.undefinedSymbols.isEmpty then .error "undefined symbols" else .ok x
+  | .ok x _ => .ok x
```
seanmcl, line 480: I spent some time trying to write `MonadLift (StM a) Err` so we didn't need this, but got stuff like "instance does not provide concrete values for (semi-)out-params". Looked at https://lean-lang.org/doc/reference/latest/Type-Classes/Instance-Synthesis/ but couldn't figure it out.

KLR/Trace/Python.lean
```diff
@@ -417,25 +417,62 @@ functions with dummy implementations, e.g., nki.language.add is defined as:
 
   def add(x,y): pass
 
-in some versions of the code. We do not want this to shadow a built-in
-definition of add. If we have an internal definition, we will use this over
-anything found during parsing.
+in the official NKI API, which the user may have imported. We do not want this
```
seanmcl, line 420: typo/grammar

interop/test/apis.py
```diff
@@ -0,0 +1,17 @@
...
+"""
+import types
+
+# Define APIs needed by test cases. We don't need the "real" versions
+# of these modules, we just want their names to be recognized as
+# modules so accesses to their contents will show up as undefined
+# names. The lean code can then deal with all of the fully-qualified
+# undefined names, e.g. `nki.isa.tensor_scalar`.
+
+np = numpy = types.ModuleType('numpy')
```
seanmcl, line 13: Seems weird to me to privilege numpy in this way. Not sure why we wouldn't ask people to import it. what about ml_dtypes and such?

.gitignore
```diff
@@ -4,3 +4,4 @@ __pycache__/
 *.dSYM
 .vscode/
 /.wheel/
+*.json
```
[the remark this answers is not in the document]
govereau: I will remove it. Better to err on the cautious side.

KLR/Python.lean
govereau: Same. I think we need to specify some `outParam`s, but it wasn't obvious, to me, what we need to do.

interop/test/apis.py
govereau: Yes, I agree. I have fixed this in a a follow-up patch.

govereau: Right now, the only use of `numpy` we have is to name ALU operations. e.g.
```
tensor_scalar(t, np.add)
tensor_scalar(t, nl.add)
```
Neither of these really makes sense. The first because there are ALU operations with no `numpy` equivalent, and the second because `nl.add` is a function on tensors not elements of tensors. The NKI API is going to be fixed to have an enumeration in `nki.isa` of the allowed ALU ops. For now, I wanted to `numpy` variants to work like the future API versions will work.

seanmcl: ok I'll just approve and we can revisit. A comment would be great.

seanmcl (approved):

PR #62, opened 2025-02-17
feat: add nki-to-klr command to the CLI

govereau:


Main.lean
```diff
@@ -245,6 +278,7 @@ def klrCmd : Cmd := `[Cli|
     compileCmd;
     gatherCmd;
     nkiToBIRCmd;
+    nkiToKLRCmd;
```
seanmcl, line 281: How is this different from `traceCmd`?

govereau: Trace command takes a json file, this one takes a python path. Same as compile vs nki-to-bir.

seanmcl (approved):

PR #64, opened 2025-02-20
WIP: refs

govereau:
Just a draft for comment. This change would simplify the types and monads, but requires the use of ST.Ref.

### DIFF
--- a/KLR/Trace/Types.lean
+++ b/KLR/Trace/Types.lean
@@ -230,7 +230,39 @@ def contains (s : State) (n : Name) : Bool :=
 
 end State
 
+instance : Inhabited State where
+  default := {}
+
 abbrev TraceM := StM State
+
+/-
+We can eliminate the Tracer Monad and the mutual block
+for Object and Term by using a ST.Ref. The ref is consistent
+with our state threading due to the definitions in KLR.Util.
+This version would not support running multiple, independent
+traces at the same time, because there is only one environment.
+
+Note: This Ref is also useful to implement mutable objects, which
+we may need at some point. Basically:
+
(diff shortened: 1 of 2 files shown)


KLR/Trace/Types.lean
```diff
@@ -230,8 +230,40 @@ def contains (s : State) (n : Name) : Bool :=
...
+we may need at some point. Basically:
+
+  instance [Obj a] : Obj (Ref a) where  -- See builtin.lean
+    ...
+
+Would allow using a `Ref` as the internal state of a builtin object.
+-/
+
+abbrev Ref := ST.Ref State
+initialize refEnv : Ref Env <- do
```
seanmcl, line 255: What does `initialize` do? I see [stuff like this](https://github.com/leanprover-community/mathlib4/blob/55df44f2c65061f8b028142fcef61830a2e8a82d/Mathlib/Tactic/Simps/NotationClass.lean#L104) in mathlib but can't find any documentation.

I also don't understand the syntax `Ref Env <-`.

govereau: It means: run this IO computation before Main and assign the result to `refEnv`.

govereau: It's a global variable.

seanmcl: I see, thanks. What about the type `Ref Env <-`? Namely the <-. I have no idea what that is.

govereau: It's is just weird syntax. I just think of it as turning into a `let refEnv : Ref Env <- ...` inside of some big IO computation somewhere with all the other `initialize` statements.

PR #68, opened 2025-03-03
chore: cleanup API and docs

govereau:
Cleanup the Python API and CLI functions and add documentation to match.


Main.lean
```diff
@@ -99,101 +100,148 @@ def gatherStr (moduleFileName kernelFunctionName : String) (klrPythonModuleDir :
...
       }
+      if output.exitCode != 0 then
+        IO.println output.stderr
+        IO.throwServerError "error gathering kernel"
       dbg $ "stdout: " ++ output.stdout
       dbg $ "stderr: " ++ output.stderr
       return output.stdout
   IO.throwServerError "could not find gather program"
 
+private def getOutfile (ext : String) (p : Parsed) :=
```
seanmcl, line 111: Could we please annotate the output of all defs? Seems like this should be an `Err`? but it looks like a `FilePath`.

interop/klr/__main__.py
```diff
@@ -0,0 +1,16 @@
...
+
+import copy
+import importlib.resources as res
+import sys
+import subprocess
+
+from .kernel import find_klr
+
+# call KLR binary
+args = copy.copy(sys.argv)
```
seanmcl, line 13: Maybe say something about trying to make this code thread safe or something. I was puzzled by the `copy`. Is the Main binary thread safe?

interop/klr/kernel.py
```diff
@@ -0,0 +1,77 @@
...
+  return d
+
+def find_klr():
+  # For development, pick up the klr binary from the project dir
+  project_root = up(os.path.abspath(__file__), 2)
+  bin = project_root + '/bin/klr'
+  if not os.path.isfile(bin):
+    # For regular pip users, pick up the klr from the wheel. While the type of `bin` here is
+    # PosixPath rather than string, they both work as the first argument to subprocess.run
+    bin = files('klr').joinpath('../bin/klr')
```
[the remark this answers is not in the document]
seanmcl: There is something wrong then. Should work no matter where you run it from. This suggests that it only works when you run it from some specific path? Please mark this as FIXME in a comment.

interop/klr/kernel.py
```diff
@@ -0,0 +1,77 @@
+# Copyright (c) 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
seanmcl, line 1: Would be good to use more type annotations in this file, especially the return types.

lakefile.lean
```diff
@@ -8,7 +8,7 @@ lean_lib "KLR" where
 
 @[default_target]
 lean_exe "klr" where
-  defaultFacets := #[LeanLib.staticFacet]
+  nativeFacets := fun _ => #[Module.oFacet]
```
seanmcl, line 11: What does this do?

KLR/Trace/FromNKI.lean
```diff
@@ -49,39 +52,55 @@ instance : FromNKI Expr where
     | .store _ _ _ => throw "store cannot be converted to KLR in this context"
     | .expr e _    => return e
 
+instance : Coe Bool Term :=
```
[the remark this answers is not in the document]
govereau: This wasn't supposed to be in this PR, slipped by. I was experimenting with some macros and this was helping a bit. I am going to revert this and if the macro stuff is useful reintroduce that in a separate PR.

KLR/Trace/FromNKI.lean
```diff
@@ -49,39 +52,55 @@ instance : FromNKI Expr where
...
+  ⟨ fun b => .expr (.const (.bool b)) .bool ⟩
+
 instance : FromNKI Bool where
   fromNKI?
-    | .expr (.const (.bool b)) .bool => return b
+    | (b : Bool) => return b
```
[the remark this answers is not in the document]
govereau: Yeah, actually this has a hidden side-effect, which is the type has to be correct. The prior could use `_` for the type which is probably better (I think the other cases do not require the type to be correct).

Main.lean
govereau: Good call.

interop/klr/kernel.py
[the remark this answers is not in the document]
govereau: Will add a comment. Also here for reference, we should probably be using the scripts directory and `sysconfig` here?
https://packaging.python.org/en/latest/specifications/binary-distribution-format/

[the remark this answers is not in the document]
govereau: Note, I couldn't figure out how to generate a wheel with a `klr-0.0.7.data/scripts` directory inside other than manually using `wheel3`.

lakefile.lean
govereau: Ah, actually the previous one doesn't do anything: it is ignored because that facet only applies to libraries. The new one adds a linker option on OSX which marks all symbols as "local". This doesn't have any noticeable effect, but it allows `strip` to remove them. I tried it, and `strip`ing the binary saves about 10% in size. Not sure if we want to do this, but it is an option now.

interop/klr/__main__.py
govereau: I don't know if this is required. It just seemed "wrong" to modify a global variable in another module. But, I guess it's python so probably fine?

govereau: Maybe something like `[find_klr] + sys.argv[1:]` is more clear. Will change.

interop/klr/kernel.py
govereau: OK. Python 3.8 support is a bit annoying, but can add some types.

seanmcl (approved):

PR #69, opened 2025-03-05
refactor: simplify KLR based on new loop design

govereau:
This change removes IndexExpr and adds a hardware-based access method (Access.pattern). This change is motivated by the new loop semantics which does not include an explicit affine loop. A future change will add a fully dynamic loop and related access forms.

### DIFF
--- a/KLR/Core/Basic.lean
+++ b/KLR/Core/Basic.lean
@@ -27,48 +27,44 @@ structure TensorName where
   memory: Memory
   deriving Repr, BEq
 
-inductive Const where
-  | none
+-- Basic indexing elements: integers and slices
+inductive Index where
+  | coord (e : Int)
+  | slice (l u step : Option Int)
+  deriving Repr, BEq
+
+-- Access pattern elements
+structure APPair where
+  step : Int := 1
+  num : Nat := 1
+  deriving Repr, BEq
+
+-- Tensor access: whole tensor (simple), basic indexing, or access pattern
+-- TODO: add advanced indexing (tensor indirect)
+inductive Access where
+  | simple (t : TensorName)
+  | basic (t : TensorName) (indexes : List Index)
+  | pattern (t : TensorName) (offset : Nat) (aps : List APPair)
+  deriving Repr, BEq
+
+inductive Value where
+  | var (x : String)
   | bool (value : Bool)
   | int (value : Int)
   | float (value : Float)
-  | string (value : String)
+  | access (a : Access)
   deriving Repr, BEq
 
--- This corresponds to the "Quasi-Affine Expressions" in Neuron.
--- Note, `floor` is the usual integer division.
-inductive IndexExpr where
-  | var (name : String)
-  | int (i : Int)
-  | neg (expr : IndexExpr)
-  | add (left right : IndexExpr)
-  | mul (scalar : Int) (expr : IndexExpr)
-  | floor (expr : IndexExpr) (scalar : Int)
-  | ceil (expr : IndexExpr) (scalar : Int)
-  | mod (expr : IndexExpr) (scalar : Int)
-  deriving Repr, BEq
-
--- Note: `np.newindex` is represented as `(.coord none)`
-inductive Index where
-  | ellipsis
-  | coord (e : Option IndexExpr)
-  | slice (l u step : Option IndexExpr)
-  deriving Repr, BEq
-
+-- TODO: Expressions are trivial right now, waiting on dynamic loops
 inductive Expr where
-  | var (x : String)
-  | const (c : Const)
-  | tensor (t : TensorName)
-  | access (t : Expr) (ix : List Index)
-  | operator (op : Operator)
-  | call (f : Expr) (args : List Expr) (kwargs : List (String × Expr))
+  | value (v : Value)
+  | call (op : Operator) (args : List Value)
   deriving Repr, BEq
 
 inductive Stmt where
-  | ret (v : Expr)
-  | store (t : TensorName) (ix : List Index) (e : Expr)
+  | ret (v : Value)
   | assign (x : String) (e : Expr)
-  | loop (x : String) (l u step : IndexExpr) (body : List Stmt)
+  | store (dst : Access) (e : Expr)
   deriving Repr, BEq
 
 structure Kernel where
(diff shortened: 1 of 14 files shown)


KLR/Core/Pretty.lean
```diff
@@ -19,106 +19,80 @@ The format here is just for ease of debugging, feel free to modify as you wish.
...
-instance : ToFormat Index      where format := ppIndex
-instance : ToFormat Operator   where format := ppOperator
-instance : ToFormat Expr       where format := ppExpr
-instance : ToFormat Stmt       where format := ppStmt
-instance : ToFormat Kernel     where format := ppKernel
+  t.name ++ sqArgs [ format t.dtype, args t.shape, format t.memory ]
+
+instance : ToFormat Kernel where
+  format k :=
+    let lines l := Format.joinSep l "\n"
```
seanmcl, line 90: Usually you don't explicitly add \n characters in a formatter. Nothing wrong with it per se, but I'd fear in more complex formatters it wouldn't be able to calculate the line length correctly during layout. Did you try adding `Format.line` here? The docs say that that may not break, but it breaks in Haskell.

KLR/Trace/Basic.lean
```diff
@@ -14,17 +14,18 @@ import KLR.Trace.NKI
...
 -- Python-like rules for conversion to boolean
-def isTrue : Const -> Bool
-  | .none     => false
+def isTrue : Value -> Bool
+  | .var _    => true
   | .bool b   => b
   | .int i    => i != 0
   | .float f  => f != 0.0
-  | .string s => s != ""
+  | .access _ => true
```
seanmcl, line 25: Why true on access? Why do we need python conventions in the core language?

KLR/Trace/Basic.lean
```diff
@@ -14,17 +14,18 @@ import KLR.Trace.NKI
...
 
 -- Python-like rules for conversion to integer
+/-
```
seanmcl, line 28: maybe kill commented code, or say why it's sticking around

KLR/Trace/Python.lean
```diff
@@ -44,42 +44,38 @@ these are syntax errors in python. Note, we do not support nested tuples or
...
+  | [], [] => return []
+  | ds, [] => return ds.map slice
+  | [], _  => throw "too many indexes for shape"
+  | d :: ds, t :: ts =>
+    match t with
+    | .none => return slice d :: (<- toIndex ds ts)
+    | .ellipsis =>
+        if ds.length == ts.length
+        then toIndex ds ts
+        else return slice d :: (<- toIndex (d::ds) ts)
```
seanmcl, line 63: This only lets the ellipsis consume a single dimension. It needs to be able to consume anything "in the middle"

```
x = np.arange(16).reshape(2,2,2,2)
x[0,...,0]Out[5]:
array([[0, 2], [4, 6]])
```

KLR/Trace/Python.lean
```diff
@@ -44,42 +44,38 @@ these are syntax errors in python. Note, we do not support nested tuples or
...
+    | .slice x y z => return .slice x y z :: (<- toIndex ds ts)
+    | .tuple _ | .list  _ => throw "nested tuple/list indexes not supported"
+    | t => return .coord (<- fromNKI? t) :: (<- toIndex ds ts)
```
seanmcl, line 66: What happens to d in this case? Seems like you'd need it.

KLR/Core/Basic.lean
```diff
@@ -27,48 +27,44 @@ structure TensorName where
...
-  | int (i : Int)
-  | neg (expr : IndexExpr)
-  | add (left right : IndexExpr)
-  | mul (scalar : Int) (expr : IndexExpr)
-  | floor (expr : IndexExpr) (scalar : Int)
-  | ceil (expr : IndexExpr) (scalar : Int)
-  | mod (expr : IndexExpr) (scalar : Int)
+-- Access pattern elements
+structure APPair where
+  step : Int := 1
```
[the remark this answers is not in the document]
govereau: As far as I can tell, this is not required to be non-zero.

KLR/Core/Basic.lean
```diff
@@ -27,48 +27,44 @@ structure TensorName where
...
-inductive Index where
-  | ellipsis
-  | coord (e : Option IndexExpr)
-  | slice (l u step : Option IndexExpr)
+-- Tensor access: whole tensor (simple), basic indexing, or access pattern
+-- TODO: add advanced indexing (tensor indirect)
+inductive Access where
+  | simple (t : TensorName)
+  | basic (t : TensorName) (indexes : List Index)
+  | pattern (t : TensorName) (offset : Nat) (aps : List APPair)
```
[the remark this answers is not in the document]
govereau: Will add some examples, agree it is opaque.

KLR/Core/Pretty.lean
govereau: I didn't try that. I will see what it does.

KLR/Trace/Basic.lean
govereau: This code is in Trace/Basic, it is only for the Python code. It is defined on Value because a Term can have a Value inside it. This case seems like a user error though...
```
if t[...]: # ?? what use could this be ??
```

KLR/Trace/Basic.lean
govereau: I will add a comment, I was keeping it around to implement `int(e)` if we decide to support this.

KLR/Trace/Python.lean
govereau: This is the case of an integer: `t[1]`. Are you saying I should check that the integer is in range of `d`?

KLR/Trace/Python.lean
seanmcl: t is the ellipsis term though. Maybe I'm confused.

KLR/Core/Basic.lean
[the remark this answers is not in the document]
govereau: I think you can, for example, broadcast with something like:
```
t[0:10,0:10] + s[[ (1,10), (0,10) ]]
```

KLR/Trace/Python.lean
govereau: Ah! Yes, you are right, I made a mistake. Nice catch.

KLR/Core/Pretty.lean
govereau: Yup, line works.

KLR/Trace/Python.lean
```diff
@@ -88,17 +96,24 @@ def list_access (name : String) (l : List Term) : Term -> Err Term
...
+  | .source _
+  | .none
   | .ellipsis
   | .slice ..
   | .store .. => throw "subscript not supported"
-  | .expr e _ => return .expr (.access e (<- termToIndex t.type i)) (.obj "object".toName)
+  | .string _ => throw "string subscript not implemented"
+  | .tuple l => list_access "list" l i
+  | .list l => list_access "tuple" l i
+  | t@(.expr ..) => do
```
seanmcl, line 112: You don't need to bind `t` here.

interop/test/test_basic.py
```diff
@@ -51,7 +51,7 @@ def expr_list(t):
 
 def expr_subscript(t):
   t[1]
-  t[1,2,3]
+  #t[1,2,3]
```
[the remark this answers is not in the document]
govereau: This is actually an improvement. The test provides a 1d tensor, and this is a legit error.
Need to go and write up lots of passing and failing tests.

KLR/Trace/Python.lean
govereau: Actually, the `fromNKI` call uses `t`.

govereau: Ah, but perhaps you mean I don't need the `.expr ..` part. Will remove.

interop/test/test_basic.py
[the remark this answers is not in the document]
seanmcl: Oh neat!

seanmcl (approved):
Looking forward to this! Seems more `eval`able

PR #71, opened 2025-03-06
feat: add memory addresses to KLR and tracing

govereau:
This is the first part of the implementation for the new allocation API in NKI. This new API allows users to create memory regions, and then attach tensors to these regions. This change introduces memory regions; attaching tensors will be done in a follow up change.

### DIFF
--- a/KLR/Core/Basic.lean
+++ b/KLR/Core/Basic.lean
@@ -4,6 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Paul Govereau, Sean McLaughlin
 -/
 import KLR.Core.Operators
+import KLR.Util
 
 /-!
 # Abstract syntax of Core NKL language
@@ -14,17 +15,67 @@ portable format, a.k.a. Kernel Language Representation (KLR).
 namespace KLR.Core
 
 /-
-A TensorName is essentially a typed variable, where the type must be a tensor
-type. This only refers to dynamic (run-time) tensors, not trace-time tensors.
+A tensor shape is a list of the sizes of each dimension of the tensor. By
+convention, the first dimension is always the "partition" dimension, and the
+remaining dimensions are "free" dimensions. When laid out in memory, the
+partition dimension will correspond to the memory partition.
 -/
 
-abbrev Shape := List Nat
+structure Shape where
+  parDim : Nat
+  freeDims : List Nat
+  deriving Repr, BEq, Inhabited
+
+def Shape.toList (shape : Shape) : List Nat :=
+  shape.parDim :: shape.freeDims
+
+def Shape.fromList : List Nat -> Err Shape
+  | [] => throw "invalid shape"
+  | p :: f => return ⟨ p, f ⟩
+
+instance : ToString Shape where
+  toString shape := toString shape.toList
+
+def Shape.freeElements (shape : Shape) : Nat :=
+  shape.freeDims.foldl (. * .) 1
+
+/-
+An Address represents a region of memory. Each address has a memory type, and
+the starting location and size of the memory region. The memory regions are
+always two-dimensional and the start and size are expressed in bytes. The
+starting location can be omitted to have the compiler choose this for you.
+
+An address may have a parent, in which case the start is relative to the
+parent's start location. This allows the user to declare a memory region with
+no start address, but containing subregions with specific relative positions.
+
+The Address structure is represented as a "pointer" term during tracing.
+-/
+
+structure Address where
+  memory : Memory
+  size : Nat × Nat
+  start : Option Nat × Option Nat := (none, none)
+  parent : Option Address := none
+  deriving Repr, BEq
+
+def Address.defaultSize (shape : Shape) (dtype : Dtype) : (Nat × Nat) :=
+  (shape.parDim * dtype.size, shape.freeElements * dtype.size)
+
+/-
+TensorName represents a tensor in memory at runtime. Each runtime tensor has a
+dtype, shape, and address. The address size must be large enough to hold the
+tensor.
+-/
 
 structure TensorName where
-  name  : String
-  dtype : Dtype
-  shape : Shape
-  memory: Memory
+  name    : String
+  dtype   : Dtype
+  shape   : Shape
+  address : Address := {
+    memory := .sbuf
+    size   := Address.defaultSize shape dtype
+  }
   deriving Repr, BEq
 
 /--
@@ -39,14 +90,28 @@ inductive Index where
   | slice (l u step : Option Int)
   deriving Repr, BEq
 
+-- Compute the number of elements an index represents
+-- Note: this assumes Index is well-formed w.r.t dim
+-- TODO: pass in proof that this is well-formed
+def Index.size (dim : Nat) : Index -> Nat
+ | .coord _ => 1
+ | .slice l u s =>
+     let abs (i : Int) : Nat := if i < 0 then (-i).toNat else i.toNat
+     let l := l.getD 0
+     let u := u.getD dim
+     let s := s.getD 1
+     let l := if l < 0 then dim + l else l
+     let u := if u < 0 then dim + u else u
+     let s := if s < 0 then -s else s
+     (abs (u - l) / s).toNat
+
 /--
 Access pattern elements.
 
-These are used for HW-native indexing which consists of an offset and a list
-of access pattern pairs. Each pair specifies a step size and the number of
-steps to take. The first pair corresponds to the partition dimension and
-changes the slowest; the last pair changes the fastest. For example, in the
-list of pairs:
+These are used for HW-native indexing which consists of an offset and a list of
+access pattern pairs. Each pair specifies a step size and the number of steps
+to take. The first changes the slowest, and the last pair changes the fastest.
+For example, in the list of pairs:
 
   [ (3,2), (1,3) ]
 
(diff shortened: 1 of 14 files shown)


KLR/Core/Basic.lean
```diff
@@ -14,17 +14,57 @@ portable format, a.k.a. Kernel Language Representation (KLR).
...
+A tensor shape is a list of the sizes of each dimension of the tensor. By
+convention, the first dimension is always the "partition" dimension, and the
+remaining dimensions are "free" dimensions. When laid out in memory, the
+partition dimension will correspond to the memory partition.
 -/
 
 abbrev Shape := List Nat
 
+def Shape.partitionElements : Shape -> Nat
+ | [] => 0
```
seanmcl, line 46: An empty shape in numpy is an array scalar. Is that a thing here? (related to former comment)

KLR/Core/Basic.lean
```diff
@@ -14,17 +14,57 @@ portable format, a.k.a. Kernel Language Representation (KLR).
...
+
 structure TensorName where
-  name  : String
-  dtype : Dtype
-  shape : Shape
-  memory: Memory
+  name    : String
+  dtype   : Dtype
+  shape   : Shape
+  address : Address := {
```
seanmcl, line 63: Should this be option? When I make a new tensor in NKI and don't specify the storage, when is it decided where it will live?

KLR/Trace.lean
```diff
@@ -14,7 +14,19 @@ import KLR.Trace.Numpy
 
 namespace KLR.Trace
 
-def globalEnv := PythonEnv ++ NKIEnv ++ NumpyEnv
+-- Keywords recognized by the tracer (KLR keywords)
+def keywords : List (Name × Term) :=
```
seanmcl, line 18: Why use Name here as opposed to String? Are there dotted paths?

KLR/Trace.lean
```diff
@@ -14,7 +14,19 @@ import KLR.Trace.Numpy
...
+  let ptr s memory size := (Lean.Name.mkStr1 s, Term.pointer { memory, size })
+  let const s := Builtin.const_var (.mkStr1 s)
+  let int s := Builtin.const_int (.mkStr1 s)
+  [ int "arch" 2
+  , const "hbm"
+  , ptr "sbuf" .sbuf (128, 0x30000)
```
seanmcl, line 24: Link to docs would be nice here.

KLR/Trace/Python.lean
```diff
@@ -96,6 +96,92 @@ def list_access (name : String) (l : List Term) : Term -> Err Term
       else throw "index out of bounds"
   |_ => throw s!"{name} indicies must be integers of slices"
 
+/-
+Access to pointer types (a.k.a. Address)
+NKI users can define memory regions by using slices on other memory regions.
+Initially, the regions `hbm` and `psum` are defined. For example:
+
+  ptr = hbm[0:32, 0:512]   # memory region in HBM
```
seanmcl, line 104: Can we be consistent about using "hbm" or "sbuf" (Core/Operators.lean)? Or is there a good reason to have them seperate?

KLR/Trace/Tensor.lean
```diff
@@ -65,8 +65,12 @@ def declare (tag : String)
     name := tname
     dtype := dtype
     shape := shape
-    memory := memory
+    address := {
+        memory := memory
+        start  := (none, none)
```
seanmcl, line 70: This is the default

KLR/Trace/Tensor.lean
```diff
@@ -65,8 +65,12 @@ def declare (tag : String)
...
+        size   := (shape.partitionElements * dtype.size, shape.freeElements * dtype.size)
```
seanmcl, line 71: This longish expression appears at least twice. Maybe define `Address.defaultSize (shape : Shape) (itemsize : Nat)`

KLR/Trace/Python.lean
```diff
@@ -96,6 +96,92 @@ def list_access (name : String) (l : List Term) : Term -> Err Term
...
+  ptr2 = ptr[:, :256]      # left half of region
+  ptr3 = ptr[:, 256:]      # right half of region
+-/
+
+def pointerAccess (addr : Core.Address) (i : Term) : Err Term := do
+  let chkPdim (p : Int) : Err Nat := do
+    if p != 0 && p != 32 && p != 64 && p != 96 then
+      throw "partition dimension must be 0, 32, 64, or 96"
```
seanmcl, line 112: Docs would be great here if they are public. And they should be if this is a hard constraint.

KLR/Trace/Python.lean
```diff
@@ -96,6 +96,92 @@ def list_access (name : String) (l : List Term) : Term -> Err Term
...
+    if p > addr.size.fst then
```
seanmcl, line 113: This looks wrong to me but maybe it's correct. The size is the partition dimension times the item size, and I'd think e.g. below where you are checking this against an index term that this should be just the dimension. I wonder if we should have a type that keeps those two numbers apart if this we are going to need to use both variants in the same function body.

KLR/Core/Basic.lean
```diff
@@ -14,17 +14,57 @@ portable format, a.k.a. Kernel Language Representation (KLR).
...
+An address may have a parent, in which case the start is relative to the
+parent's start location. This allows the user to declare a memory region with
+no start address, but containing subregions with specific relative positions.
+
+The Address structure is represented as a "pointer" term during tracing.
+-/
+
+structure Address where
+  memory : Memory
+  size : Nat × Nat
```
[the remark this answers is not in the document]
govereau: I do not know. The hbm is also 1d. Probably, there will need to be 3 kinds of addresses (1d, 2d, and register). This is just the "allocation API" which only deals with the sbuf and psum memories. I will comment on this in the PR.

KLR/Core/Basic.lean
```diff
@@ -14,17 +14,57 @@ portable format, a.k.a. Kernel Language Representation (KLR).
...
+  deriving Repr, BEq
+
+/-
+A tensor shape is a list of the sizes of each dimension of the tensor. By
+convention, the first dimension is always the "partition" dimension, and the
+remaining dimensions are "free" dimensions. When laid out in memory, the
+partition dimension will correspond to the memory partition.
 -/
 
 abbrev Shape := List Nat
```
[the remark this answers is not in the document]
govereau: That is a good call; actually makes a lot of sense here.
Is there a standard non-empty list, or do we roll our own?

KLR/Core/Basic.lean
govereau: It is not, I think on HW you always have a partition dimension. More reason to change `Shape`

KLR/Core/Basic.lean
govereau: It shouldn't be an option. I guess I am using `Address` for two slightly different things: the memory (`sbuf` and `psum` keywords), which have a `none` here and regions of the memories which cannot have a `none`. Hmm.

KLR/Trace.lean
govereau: There are dotted paths for the environment in general. Keywords will always be plain strings, but they still have to be names to go in the environment.

KLR/Trace.lean
govereau: I couldn't find a public doc. I will look again.

KLR/Core/Basic.lean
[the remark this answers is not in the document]
seanmcl: I don't see one. Here's our old one. https://github.com/leanprover/TensorLib/commit/6e07a38c8ea01a7ce930791f25c4b4234f3a1156

KLR/Trace/Python.lean
govereau: Oh, that is a mistake, should say `sbuf`. The `dram` came from BIR, but I will change it to `hbm`.

KLR/Trace/Tensor.lean
govereau: The default has `sbuf` and the memory, which may not be the case here, however I should use the default. Will update.

KLR/Trace/Tensor.lean
govereau: Will do.

KLR/Trace/Python.lean
govereau: Found one.

KLR/Trace/Python.lean
govereau: I think it is correct, but you are right that the first element is a logical partition, and the second element is a count of bytes. Double checking, there is another constraint that the free dimension has to be even. Regardless, a more structured type would make this better. I will try something new here.

KLR/Core/Basic.lean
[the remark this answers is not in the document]
govereau: I made Shape a non-empty list, the PR is getting big.

KLR/Core/Basic.lean
```diff
@@ -39,14 +90,26 @@ inductive Index where
   | slice (l u step : Option Int)
   deriving Repr, BEq
 
+-- Compute the number of elements an index represents
+-- Note: this assumes Index is well-formed w.r.t dim
+def Index.size (dim : Nat) : Index -> Nat
```
seanmcl, line 95: This doesn't work when going backwards through the array. This definition always returns 0.

```
In [21]: np.arange(16)[8:3:-2]
Out[21]: array([8, 6, 4])
```

KLR/Core/Basic.lean
```diff
@@ -62,14 +125,58 @@ structure APPair where
...
+With a meaning of:
+
+  forall p < parNum, x < num1, y < num2, ...
+    offset + p + x * step1 + y * step2 + ...
+
+The elements generated above are multiplied by the dtype size of the tensor to
+get the final memory addresses.
+-/
+
+structure AccessPattern where
```
seanmcl, line 150: Ooh this is nicer.

KLR/Core/Basic.lean
```diff
@@ -62,14 +125,58 @@ structure APPair where
...
+  | simple t | basic t .. | pattern t .. => t
+
+-- Note: this assumes Access is well-formed
+def Access.shape : Access -> Shape
+  | .simple t => t.shape
+  | .pattern _ ap => ap.shape
+  | .basic t [] => t.shape   -- Shouldn't happen, but OK
+  | .basic t (p::l) =>
+      Shape.mk (p.size t.shape.parDim)
+               ((t.shape.freeDims.zip l).map fun (d,i) => i.size d)
```
seanmcl, line 177: This zip drops elements when they overlap. Maybe return Err here or make it private and check all callers with a comment that the lengths are assumed to be the same?

KLR/Core/FromToJson.lean
```diff
@@ -61,20 +61,39 @@ instance : FromJson Engine where
   | .str s => throw s!"unknown engine type {s}"
   | _ => throw "expecting engine type"
 
+instance : ToJson Memory where
+  toJson
+  | .hbm => "dram"
```
seanmcl, line 66: Still not grokking the discrepancy. Why can't we name them the same?

govereau: The issue is that the ToJson instance is being used to generate BIR. Maybe we should introduce a new class `ToBIR` to separate these two ideas?

KLR/Core/Basic.lean
govereau: Ah, darn. Thanks.

KLR/Core/Basic.lean
govereau: I want to keep this pure, and require that `Access` is always well-formed. I am going to do this in a follow-up PR (this one is already pretty big). For now, I will insert a `panic` and include a comment referring to the future work.

KLR/Core/FromToJson.lean
seanmcl: Maybe we name the ToJson instances for BIR, put them in a namespace? Do we use ToJson for anything else?

KLR/Core/Basic.lean
seanmcl: I'll approve but could you comment or panic?

seanmcl (approved):

KLR/Core/FromToJson.lean
govereau: It is used by the CLI, but this is not strictly necessary.

seanmcl: Seems having multiple instances for the same type is tricky. My intuitions about how they work are wrong; they don't seem to respect namespace scope. So if we have 2 for the same type, which seems useful here, let's just name the top-most class instance I guess and call it directly?

PR #72, opened 2025-03-09
feat: implement memory views

govereau:
This is the second part of the new allocation API. This change adds memory views as a mechanism for declaring tensors. Included is also a change to builtin terms which allows them to carry a "self" term. This change is useful for implementing object-like methods such as the memory view function.


KLR/Trace/Basic.lean
```diff
@@ -5,6 +5,7 @@ Authors: Paul Govereau, Sean McLaughlin
 -/
 import KLR.Core
 import KLR.Trace.Types
+import KLR.Trace.Builtin
```
seanmcl, line 8: sorting imports is always nice

KLR/Trace/Basic.lean
```diff
@@ -253,18 +254,47 @@ where
...
-  tuple l := .tuple $ l.map fun i => .expr (.value (.int $ .ofNat i)) .int
+  tuple (l : List (Option Nat)) : Term :=
+    Term.tuple $ l.map fun
+      | Option.none => Term.none
+      | some x => .expr (.value (.int x)) .int
 
 -- Static environment of builtins (extend as necessary)
-
-def builtinEnv : List (Name × Builtin.BuiltinFn) :=
+open Builtin
```
seanmcl, line 273: Could we limit `open`s at the beginning of a namespace or section? Otherwise it imo makes figuring out what function is being called even harder than it already is.

KLR/Trace/FromNKI.lean
```diff
@@ -53,6 +53,11 @@ instance : FromNKI Expr where
     | .pointer ..  => err "pointer"
     | .expr e _    => return e
 
+instance : FromNKI Address where
+  fromNKI?
+  | .pointer addr => return addr
+  | _ => throw "experting pointer"
```
seanmcl, line 59: Is this a compiler error/invariant violation or user error? If the later, we would want to bubble it up so maybe need to change the type of `fromNKI?`?

KLR/Trace/Python.lean
```diff
@@ -408,7 +408,14 @@ partial def expr' : Expr' -> Trace Term
       return if tst then tru else fls
   | .call f args kws => do
       match (<- expr f : Term) with
-      | .builtin n _ => do (<- builtinFn n) (<- expr ▷ args) (<- keyword expr ▷ kws)
+      | .builtin n _ self => do
+          let f <- builtinFn n
+          let args <- expr ▷ args
```
seanmcl, line 413: IMO it's hard to search for unicode symbols. Could we maybe change this to having an ascii name and putting it in the right namespace so it's still easy to call it? It's a great function, just hard to work with when you don't have an easy unicode input method, e.g. in vscode search.

seanmcl (approved):

KLR/BIR/Compile/Memory.lean
```diff
@@ -21,27 +21,27 @@ tensors have the same basic layout.
...
 
 -- TODO: just a for instance...
-def physicalShape (t : TensorName) : Shape :=
-  match t.dtype, t.shape with
+def physicalShape (t : TensorName) : List Nat :=
+  match t.dtype, t.shape.toList with
   | .float16, [x, y] => [x, y * 2]
   | .float32, [x, y] => [x, y * 4]
-  | _, _ => t.shape -- TODO incorrect
+  | _, _ => t.shape.toList -- TODO incorrect
```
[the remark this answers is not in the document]
govereau: Actually, this can be implemented in general now. Will update.

KLR/Trace/Basic.lean
govereau: Oh, sorry. Will fix.

KLR/Trace/Basic.lean
govereau: Yup.

KLR/Trace/FromNKI.lean
govereau: It is a user error, and it does get bubbled up, but that is not obvious here. Basically, `MonadLift` takes the error (eventually) up the `Trace` monad, which has a custom `MonadExcept` instance that records the source location. I will add a comment in this file to make it clear the errors are for users.

KLR/Trace/Python.lean
govereau: Agreed, I spent 20 minutes trying to figure out which `|` I needed to type the other day.
Also, it doesn't work outside a monad. I will put a note to fix these things.

PR #73, opened 2025-03-09
feat: implement NKI allocation API

govereau:
This change builds on the proir additions of pointers and memory views to implement the NKI allocation API.

### DIFF
--- a/interop/nki/isa/__init__.py
+++ b/interop/nki/isa/__init__.py
@@ -1,0 +1,7 @@
+# KLR implemetations of NKI ISA APIs
+
+def psum_raw_ptr(address, size):
+  return psum[address[0]:size[0], address[1]:size[1]]
+
+def sbuf_raw_ptr(address, size):
+  return sbuf[address[0]:size[0], address[1]:size[1]]
--- a/interop/nki/language/__init__.py
+++ b/interop/nki/language/__init__.py
@@ -1,0 +1,8 @@
+# KLR implemetations of NKI langauge APIs
+
+def ndarray(shape, dtype, buffer=None, name=''):
+  if buffer == None:
+    buffer = sbuf
+  else:
+    assert buffer.size[0] == shape[0]
+  return buffer.view(dtype, shape, name)
--- a/interop/pyproject.toml
+++ b/interop/pyproject.toml
@@ -37,7 +37,7 @@ testpaths = [
 ]
 
 [tool.setuptools]
-packages = ["bin", "klr"] # This field is tequired to have more than 1 top level packages (nki/klr). It's not beautiful but it works.
+packages = ["nki", "bin", "klr"] # This field is required to have more than 1 top level packages (nki/klr). It's not beautiful but it works.
 
 # Build a fake C file to trick the `build` command to make an arch-specific wheel
 # We need this because we hide our `klr` Lean binary in the wheel but Python tools
--- a/interop/test/test_nki_allocation.py
+++ b/interop/test/test_nki_allocation.py
@@ -1,0 +1,45 @@
+# tests of NKI allocation API
+
+import os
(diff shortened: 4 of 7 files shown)


interop/pyproject.toml
```diff
@@ -37,7 +37,7 @@ testpaths = [
 ]
 
 [tool.setuptools]
-packages = ["bin", "klr"] # This field is tequired to have more than 1 top level packages (nki/klr). It's not beautiful but it works.
+packages = ["nki", "bin", "klr"] # This field is tequired to have more than 1 top level packages (nki/klr). It's not beautiful but it works.
```
seanmcl, line 40: typo, not yours but could you pls fix?

seanmcl (approved):

interop/test/test_nki_allocation.py
```diff
@@ -0,0 +1,45 @@
...
+
+from apis import *
+
+from klr import Kernel
+
+# Success cases
+# (these functions should load and trace to KLR)
+
+def simple():
+  ptr = nisa.sbuf_raw_ptr((0,0), (32,32))
```
aws-zhehongb, line 14: we are having some tests in the internal NKI package and some tests in the external package. This is undesirable

interop/nki/isa/__init__.py
```diff
@@ -0,0 +1,7 @@
+# KLR implemetations of NKI ISA APIs
+
+def psum_raw_ptr(address, size):
+  return psum[address[0]:size[0], address[1]:size[1]]
```
aws-zhehongb, line 4: if this is python syntax, it should be `address[0]:address[0]+size[0]`

interop/nki/language/__init__.py
```diff
@@ -0,0 +1,8 @@
+# KLR implemetations of NKI langauge APIs
+
+def ndarray(shape, dtype, buffer=None, name=''):
+  if buffer == None:
+    buffer = sbuf
+  else:
+    assert buffer.size[0] == shape[0]
```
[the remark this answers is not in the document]
govereau: At the language level, `buffer` must be a pointer type.
I think we should remove the `assert`, but this is what the current RFC requires.

interop/test/test_nki_allocation.py
govereau: Agreed.

interop/nki/isa/__init__.py
govereau: It isn't python syntax :)

interop/nki/language/__init__.py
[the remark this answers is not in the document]
govereau: `sbuf.view(...)` create a minimal size pointer.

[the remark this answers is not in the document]
aws-zhehongb: line 217 of https://code.amazon.com/reviews/CR-180753105/revisions/2#/diff suggests "`buffer` parameter here is an enum to indicating the type of buffer"

interop/nki/isa/__init__.py
aws-zhehongb: we have non-python syntax in a ".py" file?

interop/nki/language/__init__.py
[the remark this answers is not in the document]
govereau: It could be changed to:
```
if buffer == nl.sbuf:
  return sbuf.view(...)
elif buffer == l.psum:
  return psum.view(...)
```

interop/nki/isa/__init__.py
govereau: Well, I guess technically is it python syntax, it is just that the syntax is interpreted differently than normal. You raise a good point though, perhaps it should be changed.

PR #74, opened 2025-03-11
chore: remove panic from Access.shape

govereau:
This change introduces well-formed conditions to the Access type, so that all Access terms are correct by construction. The Options in the Index type are also removed as these can be computed during tracing.


KLR/Core/Basic.lean
```diff
@@ -97,9 +100,6 @@ def Index.size (dim : Nat) : Index -> Nat
  | .coord _ => 1
  | .slice l u s =>
      let abs (i : Int) : Nat := if i < 0 then (-i).toNat else i.toNat
```
seanmcl, line 102: This is `i.natAbs`

KLR/Core/Basic.lean
```diff
@@ -97,9 +100,6 @@ def Index.size (dim : Nat) : Index -> Nat
...
-     let l := l.getD 0
-     let u := u.getD dim
-     let s := s.getD 1
      let l := if l < 0 then dim + l else l
```
seanmcl, line 103: `l:Nat` so `l < 0` is always false.

KLR/Core/Basic.lean
```diff
@@ -159,25 +159,45 @@ def AccessPattern.shape (ap : AccessPattern) : Shape :=
...
   | pattern (t : TensorName) (ap : AccessPattern)
-  deriving Repr, BEq
+  deriving Repr
+
+def Access.mkBasic (t : TensorName) (indexes : List Index) : Err Access := do
+  if h : t.shape.freeDims.length + 1 = indexes.length then
+    return .basic t indexes ⟨ h ⟩
+  else throw "invalid basic access pattern"
+
+instance : BEq Access where
```
seanmcl, line 182: Since proofs are irrelevant, I guess I figured deriving BEq would do the same thing?

KLR/Core/Basic.lean
```diff
@@ -159,25 +159,45 @@ def AccessPattern.shape (ap : AccessPattern) : Shape :=
...
--- TODO: require proof of well-formedness and eliminate panic
 def Access.shape : Access -> Shape
   | .simple t => t.shape
   | .pattern _ ap => ap.shape
-  | .basic t [] => t.shape   -- Shouldn't happen, but OK
-  | .basic t (p::l) =>
-      if t.shape.freeDims.length != l.length
-      then panic "invalid access pattern" else
+  | .basic t [] wf =>
+      let h : False := by
```
seanmcl, line 196: We will have a hard time proving anything about this since we will hit dependent pattern matching on the nomatch case. I had the same thing for a while in TensorLib; when I finally tried to prove something I couldn't. Showed to John and he said he always just uses a monad even if the error case is impossible. So I switched to that and everything started working again. I recommend we do the same here. If you think it'll work, maybe let's do a couple simple proofs to see how to handle the dependent matching?

KLR/Core/Basic.lean
govereau: Oh, thanks. I could find it.

KLR/Core/Basic.lean
govereau: Good call. Actually I don't need `dim` here at all.

KLR/Core/Basic.lean
govereau: I was surprised the deriving didn't work. It complained about the cases tactic failing...

KLR/Core/Basic.lean
govereau: This is my worry; I am not sure putting these `Prop`s in the KLR AST is a good idea.
I am willing to give it a try, but we should bail if it doesn't look good.

KLR/Core/Basic.lean
```diff
@@ -60,7 +61,7 @@ structure Address where
   deriving Repr, BEq
 
 def Address.defaultSize (shape : Shape) (dtype : Dtype) : (Nat × Nat) :=
-  (shape.parDim * dtype.size, shape.freeElements * dtype.size)
```
seanmcl, line 63: Yeah, these Nats are confusing.

KLR/Core/Basic.lean
seanmcl: My proposal is to bail now and return `Err Shape`

seanmcl (approved):
Don't love the dependent matching, but 2-way door.

KLR/Core/Basic.lean
govereau: There is a small bug fix here. The `parDim` shouldn't be multiplied by the dtype size. It is fixed in the latest version of this PR.

KLR/Core/Basic.lean
govereau: Yeah, I am convinced. I pushed this PR, but I am going to change it right away.

PR #75, opened 2025-03-12
chore: bump version number to v0.0.9

govereau:
It seems like good practice to bump the version right after a release.


seanmcl (commented):
we need some mechanism to push a tag

seanmcl (approved):

govereau:
> we need some mechanism to push a tag

Can we have github require an "annotated tag", which has a commit object?

PR #77, opened 2025-03-12
chore: sync BIR instructions

govereau:
This change updates the generated BIR instructions. There is a change to how the instructions are generated: we no longer emit empty structures. The Block type is also simplified to avoid mutual inductive types.


KLR/BIR/Instructions.lean
```diff
@@ -159,81 +159,75 @@ inductive BranchCompareOp where
 
 -- Instruction
 
-structure Activation extends Instruction where
+structure Activation /- extends Instruction -/ where
```
seanmcl, line 162: ?

KLR/BIR/Instructions.lean
```diff
@@ -159,81 +159,75 @@ inductive BranchCompareOp where
...
-  deriving BEq, Repr, Lean.FromJson, Lean.ToJson
+-- GetGlobalRankId (extends Instruction) has no fields
 
-structure NoOp extends Instruction where
-  deriving BEq, Repr, Lean.FromJson, Lean.ToJson
+-- NoOp (extends Instruction) has no fields
 
-structure EventSemaphore extends Instruction where
-  deriving BEq, Repr, Lean.FromJson, Lean.ToJson
+-- EventSemaphore (extends Instruction) has no fields
```
seanmcl, line 185: Obvious from the previous 2 lines?

seanmcl (approved):

KLR/BIR/Instructions.lean
govereau: I was trying to reduce compile times by using `extends` less and not generating empty structures. In the newly generated code, instead of everything extending `Instruction`, the `Instruction` type includes an `Inst`, this seems more idiomatic (for Lean) to me. I left the comments in the generated code as hints in case someone is also looking at the BIR classes.

KLR/BIR/Instructions.lean
govereau: I had the generator script emit these comments when it doesn't have anything to output because the structure is empty. Just a hint for readers trying to connect this to the BIR classes. There is no Lean type corresponding to `EventSemaphore`, because we don't need one: the inductive type `Inst` has an empty variant for this instruction, but we don't also need an empty type.

PR #78, opened 2025-03-13
chore: Make Slice a type and fix the size function

seanmcl:
Slices, even with nat bounds, are a bit too tricky to not have their own type. This PR adds the Slice type and pushes it through the uses.

### DIFF
--- a/KLR/Util.lean
+++ b/KLR/Util.lean
@@ -46,3 +46,9 @@ notation f "▷" l =>
 
 def impossible {a : Type} [h : Inhabited a] (msg : String := "") :=
   @panic a h s!"Invariant violation: {msg}"
+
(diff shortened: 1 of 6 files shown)


KLR/Util.lean
```diff
@@ -46,3 +46,9 @@ notation f "▷" l =>
 
 def impossible {a : Type} [h : Inhabited a] (msg : String := "") :=
   @panic a h s!"Invariant violation: {msg}"
+
+def get! [Inhabited a] (x : Err a) : a := match x with
```
seanmcl, line 50: We can get these (and impossible) from TensorLib when we link it, but for now I thought copying was better.

govereau (approved):
Much nicer.

PR #80, opened 2025-03-14
chore: simplify getting partition/free dimension offsets

seanmcl:


KLR/Core/Basic.lean
```diff
@@ -231,9 +232,21 @@ structure AccessPattern where
   offset : Nat := 0
   deriving Repr, BEq
 
-def AccessPattern.shape (ap : AccessPattern) : Shape :=
+namespace AccessPattern
+
+def shape (ap : AccessPattern) : Shape :=
   .mk ap.parNum $ ap.freePattern.map fun pair => pair.num
 
+def partitionByteOffset (ap : AccessPattern) : Nat := ap.tensor.address.partitionOffset.getD 0
```
seanmcl, line 240: @govereau please make sure I did this correctly.

govereau: I think this is not quite right. The partition is always a number between 0-128, there isn't a byte offset for the partition, only for the "free dimensions" within the partition(s).

KLR/Core/Basic.lean
```diff
@@ -231,9 +232,21 @@ structure AccessPattern where
...
+
+def partitionElementOffset (ap : AccessPattern) : Nat := ap.partitionByteOffset / ap.tensor.dtype.size
+
+def freeByteOffset (ap : AccessPattern) : Nat := ap.tensor.address.freeOffset.getD 0 + ap.offset
```
govereau, line 244: Yes, I think this is right. It is confusing because the APPairs are in elements and the offset is in bytes. I double checked the BIR and HW references, and seems to be the case.

KLR/Core/Basic.lean
```diff
@@ -231,9 +232,21 @@ structure AccessPattern where
...
+
+def freeElementOffset (ap : AccessPattern) : Nat := ap.freeByteOffset / ap.tensor.dtype.size
```
govereau, line 246: Here I am not so sure. I can't find anything that says the offset has to be aligned to the data type size. Probably it does, just not sure.

govereau (changes requested):
One small issue regarding partition dimension.

KLR/Core/Basic.lean
seanmcl: Ah of course. I'll call it `partitionRowOffset` or something.

KLR/Core/Basic.lean
seanmcl: I'll add a "not sure" comment.

govereau (approved):

PR #81, opened 2025-03-14
feat: first cut at eval

seanmcl:


KLR/Eval.lean
```diff
@@ -0,0 +1,293 @@
...
+def lookup (x : String) : WithEnv (Option EValue) := do
+  let env <- get
+  return env.lookup x
+
+def lookupTensor (t : String) : WithEnv Tensor := do (<- get).lookupTensor t
+
+def insertTensor (x : String) (t : Tensor) : WithEnv Unit := do
+  modifyGet fun env => ((), env.insert x (.ETensor t))
+
+def pure (x : a) : WithEnv a := StateT.pure x
```
govereau, line 56: Here is a case where you could use `export StateT(pure lift)`. (or open)

KLR/Eval.lean
```diff
@@ -0,0 +1,293 @@
...
+#guard Iterator.toList (APPairIter.make (APPair.mk (-1) 5) 10) == [(10:Int), 9, 8, 7, 6]
+#guard Iterator.toList (APPairIter.make (APPair.mk (-2) 5) 3) == [(3:Int), 1, -1, -3, -5]
+
+end APPairIter
+
+def APPairs := List APPair
+
+namespace APPairs
+
+abbrev Iter := Iterator (LEList APPairIter) (List Int)
```
govereau, line 156: Oh nice! `LEList` just works here.

KLR/Eval.lean
```diff
@@ -0,0 +1,293 @@
...
+#guard (Iterator.toList (lelist [APPair.mk 1 1, APPair.mk 1 2, APPair.mk 1 2] 0) : List (List Int))
+  == [[(0:Int), 0, 0], [0, 1, 0], [0, 0, 1], [0, 1, 1]]
+
+end APPairs
+
+def AccessPattern.lelist (p : AccessPattern) : Err (LEList APPairIter) := do
+  let parOff <- p.partitionRowOffset
+  let freeOff <- p.freeElementOffset
+  let parDimIter := APPairIter.make (APPair.mk 1 p.parNum) parOff
+  let freeIters := p.freePattern.map fun p => APPairIter.make p freeOff
```
govereau, line 180: Does every free element get `freeOff` or just the first one? It looks like we get `freeOff` multiple times?

KLR/Eval.lean
```diff
@@ -0,0 +1,293 @@
...
+  for resIndex in shapeIter do
+    let index : List Int := Iterator.peek apIter
+    let apIter' <- match Iterator.next (List Int) apIter with
+    | none => throw "iterator failed"
+    | some iter =>
+    apIter := iter
+    let index <- index.mapM fun i => if i < 0 then .error "negative index" else return i.toNat
+    let v <- t.getDimIndex index
+    res <- res.setDimIndex resIndex v
+  return res
```
govereau, line 227: Pretty dense, but the algorithm is just complex. 18 lines is not bad since you are replicating HW behavior here.

KLR/Eval.lean
```diff
@@ -0,0 +1,293 @@
...
+  let t <- lookupTensor tname.name
+  let () <- checkIOTensor tname t
+  let shape := evalShape pattern.shape
+  let mut res := TensorLib.Tensor.zeros t.dtype shape
+  let mut apIter <- AccessPattern.lelist pattern
+  let shapeIter := TensorLib.Shape.belist shape
+  if Iterator.size (List Int) apIter != Iterator.size (List Nat) shapeIter then .error "invariant violation: iterator size mismatch" else
+  for resIndex in shapeIter do
+    let index : List Int := Iterator.peek apIter
+    let apIter' <- match Iterator.next (List Int) apIter with
```
govereau, line 220: You could write a MonadLift instance to avoid match on Iterator.next?
I guess it is Option so perhaps a bit weird.

KLR/Eval.lean
```diff
@@ -0,0 +1,293 @@
...
+  -- NB: This is weird. To evaluate a return, we shove a value with the special string "result" into the environment.
+  let res <- evalValue v
+  modify fun env => env.insert "result" res
+| .assign x e => do
+  let v <- evalExpr e
+  modify fun env => env.insert x v
+| .store dst (Operator.tensorScalar (TensorScalar.mk AluOp.add n false AluOp.bypass _ _)) [.access v] => do
+  let dtype <- evalDtype v.tensor.dtype
+  let v <- read v
+  let n <- TensorLib.Tensor.arrayScalar dtype n.toLEByteArray
```
govereau, line 274: I think the HW will cast to fp32, do the add, and then cast down to the dst type. Should check though.

govereau (approved):
Looks pretty good to me.

KLR/Eval.lean
```diff
@@ -0,0 +1,293 @@
...
+    | some (.ETensor t) => checkIOTensor out t
+    | some _ => .error s!"Wrong type mapping for {out.name}"
+    | none => .error s!"Missing mapping for {out.name}"
+
+-- This differs from IntIter because step can be 0. We don't have a stop, but instead
+-- have a count
+structure APPairIter where
+  private mk::
+  pair : APPair
+  start : Int
```
govereau, line 110: Start is the offset? If so the units are bytes, but I guess you would want to divide by the dtype size for the iterator?

seanmcl: May be good to make typedefs for ElemCount vs ByteCount. We already divide by the bytes via `freeElementOffset` so start is in element-count.

KLR/Eval.lean
seanmcl: Yeah. May switch to `BEList` bc prob makes more sense from locality perspective (not that it matters that much) but it's neat that we get these annoying carry-iterators for free.

KLR/Eval.lean
seanmcl: What I intended was for each partition to start at `freeOff` elements in. I think that's the correct mental model, though I could have messed up byte vs element count here.

KLR/Eval.lean
seanmcl: ah good point.

KLR/Eval.lean
govereau: That is the right model, I am confused because if you have say a 3d tensor, then there are two free dimensions, but still just one offset into the partition. I was expecting to see one partition iterator, one APIter with an offset and one APIter without an offset.

seanmcl: Oh ug I see your point now. Yeah it's wrong.

KLR/Eval.lean
seanmcl: I'll keep an eye out for simplifications. It does some sanity checks, than walks over two iterators in tandem. I wonder if a 

```
for i, j in iter1, iter2 do
  ...
```

syntax would make this kind of code nicer. I'll play around with it. Walking over two iterators, one for the src and one for the dst, will be a common pattern.

KLR/Eval.lean
seanmcl: This is fixed.

KLR/Eval.lean
seanmcl: Ah, that's actually nicer. Done.

KLR/Eval.lean
seanmcl: Moved to FP32 throughout.

lakefile.lean
```diff
@@ -13,3 +13,5 @@ lean_exe "klr" where
 
 require Cli from git
   "https://github.com/leanprover/lean4-cli.git" @ "v4.16.0"
+
+require TensorLib from "../TensorLib"
```
seanmcl, line 17: I'll point this at a release once it's close to ready to go.

govereau (approved):
This looks good. Sorry the operators and AluOps are a bit messy, but getting cleaned up.

KLR/Eval.lean
seanmcl: No obvious solution for the syntax. I can make a pair, and maybe that's the right way to go. Will experiment in another PR, since it will impact multiple spots where we do this both here and in TensorLib.

seanmcl: [This works nicely I think](https://github.com/leanprover/TensorLib/pull/56)

PR #84, opened 2025-03-19
chore: remove the triangle notation

seanmcl:
The triangle notation was introduced to get around a limitation in Lean that couldn't prove termination when mapM was used, even if it was structurally
obvious. I didn't like the notation solution because it seemed brittle; I tried replacing the notation with a def and an abbrev, but neither worked. I still don't understand why it works, but I also don't understand why

    | .tuple es .store => return .tuple (<- es.attach.mapM fun ⟨ e, _ ⟩ => LValue e)

works but

    | .tuple es .store => return .tuple (<- es.attach.mapM fun e => LValue e.val)

does not. Since this is now required in only 2 spots (thanks likely to improvements in Lean itself) I propose we just remove the notation and do the trick until it's no longer needed.

### DIFF
--- a/KLR/Trace/Python.lean
+++ b/KLR/Trace/Python.lean
@@ -257,8 +257,8 @@ where
   lval : Expr' -> Trace Term
   | .name id .store => return .expr (.value $ .var id) (.obj "object".toName)
   | .attr _ id .store => throw s!"cannot assign to attribute {id}"
-  | .tuple l .store => return .tuple (<- LValue ▷ l)
-  | .list  l .store => return .list  (<- LValue ▷ l)
+  | .tuple es .store => return .tuple (<- es.attach.mapM fun ⟨ e, _ ⟩ => LValue e)
+  | .list es .store => return .list (<- es.attach.mapM fun ⟨ e, _ ⟩ => LValue e)
   | .subscript _ _ .store => throw "unimp subscript store"
   | _ => throw "cannot assign to expression"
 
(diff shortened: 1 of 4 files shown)


KLR/Trace/Python.lean
```diff
@@ -257,8 +257,8 @@ where
   lval : Expr' -> Trace Term
   | .name id .store => return .expr (.value $ .var id) (.obj "object".toName)
   | .attr _ id .store => throw s!"cannot assign to attribute {id}"
-  | .tuple l .store => return .tuple (<- LValue ▷ l)
-  | .list  l .store => return .list  (<- LValue ▷ l)
+  | .tuple es .store => return .tuple (<- es.attach.mapM fun ⟨ e, _ ⟩ => LValue e)
+  | .list es .store => return .list (<- es.attach.mapM fun ⟨ e, _ ⟩ => LValue e)
```
govereau, line 261: I think what is going on here is that the Lean compiler is going to try to prove termination using some tactics, and these tactics fail. However, if you introduce the proofs that all of the list elements are in the list using `attach`, then the tactic succeeds. I don't know why this is makes the tactic work, but this extra hypothesis is what is needed by the tactics.

However, If you use `fun x => ... x.fst` then the tactic needs to find the proof inside of `x`, which it fails to do. Again, I don't know why, but this makes some sense to me that the tactic might not know it needs to deconstruct `x` to make progress. If you instead write `fun  (e, _) => ...`, then the proof the tactic needs is sitting in the hypotheses, it doesn't matter that it doesn't have a name, the tactic can still find it, and prove the goal. I guess in this case the needed proof would have a meta-variable name, but the tactic doesn't care.

govereau (approved):
Nice to see Lean is getting better at automation.

KLR/Trace/Python.lean
seanmcl: FWIW, I tried defining a depth function and proving termination without the `.attach` calls. It was a lot of code and I abandoned the attempt. I didn't appreciate until now how much work the termination automation is saving us.

PR #94, opened 2025-04-21
refactor: make Python AST more amenable to extraction to C

govereau:
Small modifications to Python AST to make it easier to automatically generate equivalent C and Python versions. Changes include: replacing single constructor inductive types with structures, replacing Prod types with structures, renaming some constructors to avoid conflicts, replacing HashSet with List.

### DIFF
--- a/KLR/Python.lean
+++ b/KLR/Python.lean
@@ -48,9 +48,9 @@ inductive Ctx where
   | load | store | del
   deriving Repr
 
--- Python boolean operators
+-- Python boolean (logical) operators
 inductive BoolOp where
-  | and | or
+  | land | lor
   deriving Repr
 
 -- Python comparison operators
@@ -71,8 +71,9 @@ inductive BinOp where
   deriving Repr
 
 mutual
-inductive Expr where
-  | exprPos (expr : Expr') (pos : Pos)
+structure Expr where
+  expr : Expr'
+  pos : Pos
   deriving Repr
 
 inductive Expr' where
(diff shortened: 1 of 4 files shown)


KLR/Trace/Python.lean
```diff
@@ -251,8 +251,8 @@ dead-code elimination for simple assignments.
 
 -- Convert an expression in assignment context (an L-Value).
 -- TODO: handle subscript
-def LValue : Expr -> Trace Term
-  | .exprPos e' p => withPos p (lval e')
+partial def LValue (e : Expr) : Trace Term :=
```
seanmcl, line 254: Why `partial`?

KLR/Python.lean
```diff
@@ -190,11 +196,11 @@ instance : Repr Lean.Json where
...
-  funcs : List (String × Fun)
-  args : List Expr'
-  kwargs : List (String × Expr')
-  globals : List (String × Expr')
-  undefinedSymbols : Std.HashSet String
+  funcs : List Fun
+  args : List Expr
+  kwargs : List Keyword
+  globals : List Keyword
+  undefinedSymbols : List String
```
seanmcl, line 203: Why did this change to a list?

seanmcl (commented):
Nice, looks better.

seanmcl (approved):

KLR/Python.lean
govereau: It is because I don't have a set in C. If I was generating C++ I could generate something.

KLR/Trace/Python.lean
govereau: The mutual recursion between the structure and inductive seems to have caused the automation failure. I wasn't able to figure out how to prove it.

govereau: I will take another look at it when I fixup the tracer to use NKI syntax.

PR #95, opened 2025-04-21
feat: introduce NKI abstract syntax

govereau:
This change introduces the abstract syntax of NKI.  This change is part of a larger effort to break KLR up into smaller passes that can be more easily compared with the upcoming C++ implementation. The first of these passes will be a simplification pass that consumes the Python Core AST and produces the NKI AST (contained in this patch).

### DIFF
--- a/KLR/NKI/Basic.lean
+++ b/KLR/NKI/Basic.lean
@@ -1,0 +1,114 @@
+/-
+Copyright (c) 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Paul Govereau, Sean McLaughlin
+-/
+
+/-!
+# Abstract syntax of NKI functions
+
+This AST is produced from the Python AST by simplification.
+The two ASTs are similar. This one is the "official" abstract
+syntax of NKI.
+-/
+
+namespace KLR.NKI
+
+structure Pos where
+  line : Nat
+  column : Nat
+  deriving Repr
+
+-- Note: the python int and float types are compatible with Lean's types
+-- The str type may require conversion (to UTF8).
+inductive Value where
+  | none
+  | bool (value : Bool)
+  | int (value : Int)
+  | float (value : Float)
+  | string (value : String)
+  | ellipsis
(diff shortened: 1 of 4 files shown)


KLR/NKI/Basic.lean
```diff
@@ -0,0 +1,112 @@
...
+structure Pos where
+  line : Nat
+  column : Nat
+  deriving Repr
+
+inductive Value where
+  | none
+  | bool (value : Bool)
+  | int (value : Int)
+  | float (value : Float)
```
seanmcl, line 26: Maybe comment here that Python and Lean line up for Float? I checked for int/float which do match.

seanmcl (approved):

KLR/NKI/Basic.lean
```diff
@@ -0,0 +1,112 @@
...
+Copyright (c) 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Paul Govereau, Sean McLaughlin
+-/
+
+/-!
+# Abstract syntax of NKI functions
+
+This AST is produced from the Python AST by simplification.
+The two ASTs are similar. This one is the "official" abstract
```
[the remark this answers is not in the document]
govereau: Yes. I plan to write a new parser (eventually) to target this AST directly.

KLR/NKI/Basic.lean
```diff
@@ -0,0 +1,112 @@
...
+  line : Nat
+  column : Nat
+  deriving Repr
+
+inductive Value where
+  | none
+  | bool (value : Bool)
+  | int (value : Int)
+  | float (value : Float)
+  | string (value : String)
```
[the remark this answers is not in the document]
govereau: Good point, will add a comment.

KLR/NKI/Basic.lean
govereau: Perhaps I should use `Int64`?

seanmcl: Python ints are unbounded, right? So matches Lean iiuc.

govereau: Yes, they are, but I could check them in the parser before building the AST.
Probably I will save this change until the new parser is being built.

PR #96, opened 2025-04-21
feat: Add SHA256

seanmcl:
Will use this for NEFF packaging


govereau (approved):
Oh Great! I need this for assigning unique ids to the NKI kernels as well.

PR #97, opened 2025-04-22
feat: Add Base64

seanmcl:


KLR/Util/Base64.lean
```diff
@@ -0,0 +1,179 @@
...
+private def makeDecodeTable : Array UInt8 := Id.run do
+  let mut table := Array.mkArray 256 0
+  let mut i := 0
+  for c in encodeTable do
+    table := table.set! c.toNat i
+    i := i + 1
+  table
+
+/-- The decode table for Base64 (cached for performance) -/
+private def decodeTable : Array UInt8 := makeDecodeTable
```
govereau, line 29: What is the difference between this and `makeDecodeTable`?

govereau (approved):

KLR/Util/Base64.lean
seanmcl: Claude has a thing or two to learn about the guarantees you get from types.

PR #98, opened 2025-04-22
feat: Add basic NEFF codec

seanmcl:
Example:

    $ bin/klr neff-info add.neff
    { header := { packaging_version := 2,
              header_size := 1024,
              data_size := 30720,
              kelp_version_major := 1,
              kelp_version_minor := 0,
              kelp_version_build := #[0, 0, ...],
              num_tpb := 1,
              hash := #[],
              uuid := #[],
              name := #[],
              requested_tpb_count := 0,
              tpb_per_node := #[],
              pad := #[0, 0, ...] },
    data := #[31, 139, ...

### DIFF
--- a/KLR/NEFF/Basic.lean
+++ b/KLR/NEFF/Basic.lean
@@ -1,0 +1,203 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Paul Govereau, Sean McLaughlin
+-/
+
+/-
+NEFF: Neuron Executable File Format
+-/
+namespace KLR
+namespace NEFF
+
+private def _root_.ByteArray.appendUInt64 (ba : ByteArray) (val : UInt64) : ByteArray :=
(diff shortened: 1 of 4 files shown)


KLR/NEFF/Basic.lean
```diff
@@ -0,0 +1,203 @@
...
+    num_tpb
+    hash
+    uuid
+    name
+    requested_tpb_count
+    tpb_per_node
+    pad
+  if !header.validate then none else some header
+
+def serialize (header : Header) : ByteArray := Id.run do
```
govereau, line 118: For the ISA we will also need functions to convert to and from ByteArray. Perhaps we should define a couple of classes and instances for basic types. Setting us up to implement an automatic `ToByteArray` and `FromByteArray` deriving?

govereau (approved):

KLR/NEFF/Basic.lean
seanmcl: Yes. To get at the NEFF body I'm working on tar and gzip now. Once there, and with a couple of ISA examples under our belts, I think there will be enough examples to generalize all of the bytearray codecs.

PR #99, opened 2025-04-23
feat: add extraction for Lean types

govereau:
This change introduces a simple extractor for Lean types, along with a basic skeleton for compiling the C version of KLR.


interop/nkic/frontend.c
```diff
@@ -0,0 +1,146 @@
...
+#include "ast_nki.h"
+
+// This file defines the frontend Python extension module and the
+// Kernel type contained therein.
+
+// frontend.Kernel.__init__
+static int kernel_init(struct kernel *self, PyObject *args, PyObject *kwds) {
+  (void)kwds;
+  PyObject *f = NULL;
+  if (!PyArg_ParseTuple(args, "O", &f)) {
```
graebm, line 18: trivial: use PyArg_ParseTupleAndKeywords() as shown in [this example](https://docs.python.org/3/extending/newtypes_tutorial.html#adding-data-and-methods-to-the-basic-example) to raise error if user passed anything via `kwds`

interop/nkic/frontend.c
```diff
@@ -0,0 +1,146 @@
...
+// Custom deallocator for Kernel type
+static void kernel_dealloc(struct kernel *self) {
+  Py_XDECREF(self->f); // NULL is OK
+  Py_TYPE(self)->tp_free((PyObject *) self);
+}
+
+// frontend.Kernel.specialize
+// Provide arguments for kernel specialization
+static PyObject* kernel_specialize(struct kernel *self, PyObject *args, PyObject *kws) {
+  if (!PyTuple_Check(args) || (kws && !PyDict_Check(kws))) {
```
graebm, line 42: trivial: PyArg_ParseTupleAndKeywords() probably a better call

interop/nkic/frontend.c
```diff
@@ -0,0 +1,146 @@
...
+  (void)self;
+  (void)args;
+  return PyLong_FromLong(KLR_VERSION);
+}
+
+// frontend.deserialize
+static PyObject* deserialize(PyObject *self, PyObject *args) {
+  (void)self;
+  PyObject *ba = NULL;
+  if (!PyArg_ParseTuple(args, "Y", &ba)) {
```
graebm, line 72: Python [docs](https://docs.python.org/3/c-api/arg.html#strings-and-buffers) suggest `y*`
>  This is the recommended way to accept binary data.

interop/nkic/frontend.c
```diff
@@ -0,0 +1,146 @@
...
+  }
+  self->specialized = true;
+  return Py_None;
+}
+
+// frontend.Kernel.serialize
+static PyObject* kernel_serialize(struct kernel *self, PyObject *args) {
+  (void)args;
+  if (!self->specialized) {
+    PyErr_SetString(PyExc_Exception, "specialized must be called before serialize");
```
graebm, line 54: utterly trivial: 
```suggestion
    PyErr_SetString(PyExc_Exception, "specialize must be called before serialize");
```

interop/nkic/frontend.c
```diff
@@ -0,0 +1,146 @@
...
+
+PyMODINIT_FUNC PyInit_frontend(void) {
+  if (PyType_Ready(&KernelType) < 0)
+    return NULL;
+
+  PyObject *m = PyModule_Create(&module);
+  if (!m)
+    return NULL;
+
+  // PyModule_AddObjectRef is not available in Python 3.9
```
graebm, line 134: going back to the 3.9 version of these docs, seems like they use PyModule_AddObject() instead
https://docs.python.org/3.9/extending/newtypes_tutorial.html#adding-data-and-methods-to-the-basic-example

interop/nkic/ast_nki.h
```diff
@@ -0,0 +1,412 @@
...
+struct NKI_Pos {
+  u32 line;
+  u32 column;
+};
+
+struct NKI_Value {
+  enum NKI_Value_Tag {
+    NKI_Value_NONE,
+    NKI_Value_BOOL,
+    NKI_Value_INT,
```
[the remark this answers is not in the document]
govereau: Is it a good question. Currently NKI Ints are big ints, but I was thinking we could check that they are 32-bit when converting from Python (because the HW is only 32-bit). There is a comment in the original lean code about the ints being compatible with Python, but it is an open issue.

interop/nkic/frontend.c
govereau: Hmm. If I use `y*` will the return still be guaranteed to be a `PyByteArrayObject`? In particular, will the two calls below be guaranteed to succeed?

interop/nkic/ast_nki.h
```diff
@@ -0,0 +1,412 @@
...
+
+static inline struct NKI_Expr *mkNKI_Expr_proj(struct NKI_Expr *expr,
+                                               const char *name,
+                                               struct region *region) {
+  struct NKI_Expr *res = region_alloc(region, sizeof(*res));
+  if (!res)
+    return NULL;
+  res->expr.tag = NKI_Expr_PROJ;
+  res->expr.proj.expr = expr;
+  res->expr.proj.name = name;
```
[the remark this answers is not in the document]
govereau: I am expecting that `name` (and everything else is in the same region.

BTW: How do you get github to show the "commit suggestion" button? Super convenient.

interop/nkic/frontend.c
govereau: Ah, cool. I will change to that.

interop/nkic/frontend.c
govereau: This one is trickier: we don't know what we are expecting, we need to match these up with the user's kernel arguments. This part will come after gather is complete (which is the next step).

interop/nkic/frontend.c
govereau: I will check for keywords.

interop/nkic/frontend.c
graebm: With `y*`, you'd get a [Py_buffer](https://docs.python.org/3/c-api/buffer.html) down here in C. Advantages `y*` over `Y` are:
- it would accept more than just `bytearray` from Python. Also `bytes`, `array.array`, and more.
- it locks the underlying buffer so that the caller can subsequently use the buffer even inside a [Py_BEGIN_ALLOW_THREADS](https://docs.python.org/3/c-api/init.html#c.Py_BEGIN_ALLOW_THREADS) block without the risk of mutable data being resized or destroyed. 

Disadvantages are:
- you have to call [PyBuffer_Release()](https://docs.python.org/3/c-api/buffer.html#c.PyBuffer_Release) after you have finished processing the data (or in any early abort case).

We could always change this later. If you want to stick with PyByteArrayObject for now.

interop/nkic/ast_nki.h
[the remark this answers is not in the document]
graebm: - when adding the comment, click-drag to highlight multiple lines
- click the `+-` "Add a suggestion" button (first button in line of formatting helper buttons)
- that puts the code in your comment box
- edit it as you see fit

interop/nkic/frontend.c
govereau: OK, that looks better, but there seems to be some extra complexity around shape and striding. Let's revisit in a later PR.

interop/nkic/frontend.c
```diff
@@ -0,0 +1,146 @@
...
+PyMODINIT_FUNC PyInit_frontend(void) {
+  if (PyType_Ready(&KernelType) < 0)
+    return NULL;
+
+  PyObject *m = PyModule_Create(&module);
+  if (!m)
+    return NULL;
+
+  if (PyModule_AddObject(m, "Kernel", (PyObject*) &KernelType) < 0) {
+    Py_DECREF(m);
```
graebm, line 142: trivial: more boilerplate from the [3.9 example](https://docs.python.org/3.9/extending/newtypes_tutorial.html#adding-data-and-methods-to-the-basic-example). I guess this is why they added the new better function in 3.10
```suggestion
  Py_INCREF(&KernelType);
  if (PyModule_AddObject(m, "Kernel", (PyObject*) &KernelType) < 0) {
    Py_DECREF(&KernelType);
    Py_DECREF(m);
```

interop/nkic/frontend.c
```diff
@@ -0,0 +1,146 @@
...
+  }
+  self->specialized = true;
+  return Py_None;
+}
+
+// frontend.Kernel.serialize
+static PyObject* kernel_serialize(struct kernel *self, PyObject *args) {
+  (void)args;
+  if (!self->specialized) {
+    PyErr_SetString(PyExc_Exception, "specialize must be called before serialize");
```
graebm, line 61: utterly trivial:
```suggestion
    PyErr_SetString(PyExc_RuntimeError, "specialize must be called before serialize");
```

interop/nkic/frontend.c
govereau: I had to rewrite this in the next PR; so I will make the changes there.

interop/nkic/frontend.c
govereau: Good call. Updated in next PR.

KLR/Extract/Extract/Basic.lean
```diff
@@ -0,0 +1,121 @@
...
+  name : Name
+  type : SimpleType
+  deriving Repr
+
+inductive LeanType where
+  | prod (name : Name) (fields : List Field)
+  | sum (name : Name) (variants : List LeanType)
+  deriving Repr
+
+instance : Inhabited LeanType where
```
seanmcl, line 39: Do you care about the default? `derive Inhabited, Repr` works fine here.

KLR/Extract/Extract/Basic.lean
```diff
@@ -0,0 +1,121 @@
...
+end LeanType
+
+private def collectType : Expr -> MetaM SimpleType
+  | .const `Bool [] => return .bool
+  | .const `Nat [] => return .nat
+  | .const `Int [] => return .int
+  | .const `Float [] => return .float
+  | .const `String [] => return .string
+  | .const n [] => return .const n
+  | .app (.const `List [0]) t => return .list (<- collectType t)
```
seanmcl, line 73: what is the `[0]`?

KLR/Extract/Extract/Basic.lean
```diff
@@ -0,0 +1,121 @@
...
+-- Note: we want the list to remain in given order
+def collectLeanTypes (names : List Name) : MetaM (List LeanType) := do
+  let mut enums := []
+  let mut res := []
+  for name in names do
+    let ty <- collectLeanType name
+    if ty.isEnum then
+      enums := name :: enums
+    res := ty :: res
+  return res.reverse.map fun t => t.rewriteEnums enums
```
seanmcl, line 121: I wanted to write the marginally better `.map (rewriteEnums enums)` but it's less nice since you need to add `LeanType.rewriteEnums enums`.

KLR/Extract/Extract/C.lean
```diff
@@ -0,0 +1,234 @@
...
+    IO.println s!"static inline {ptr}"
+    IO.println s!"{fnName}({args} struct region *region)"
+    genInitBody retTy t
+  | .sum _ variants =>
+    for v in variants do
+      genMkFuns retTy v
+
+def pythonAST: MetaM (List LeanType) := do
+  collectLeanTypes [
+    `KLR.Python.Pos,
```
seanmcl, line 153: Sort?

interop/nkic/README.md
```diff
@@ -0,0 +1,51 @@
...
+| C File | Lean Source | Method |
+|-|-|-|
+| ast_python_core.h | KLR/Python.lean    | KLR/Extract/C |
+| ast_nki.h         | KLR/NKI/Basic.lean | KLR/Extract/C |
+| ast_nki.py        | KLR/NKI/Basic.lean | KLR/Extract/Python |
+
+## CPython Sources
+
+The files in the `peg_parser` directory are from the CPython sources. These
+files have been lightly modified for use in NKI.
```
seanmcl, line 20: Is it obvious how they are modified?

KLR/Extract/Extract/Basic.lean
govereau: Actually, I don't need this instance at all. I removed it.

KLR/Extract/Extract/Basic.lean
govereau: The zero is the type universe level.

KLR/Extract/Extract/C.lean
govereau: This list defines the order the definitions are output to the C file. They have to be in dependency order or the C compiler will fail.

interop/nkic/README.md
govereau: I will add more detail to the README. Mostly just marking functions "static" and deleting function we don't use.

seanmcl (approved):
Next level metaprogramming. You need to teach a class on this.

interop/nkic/ast_nki.py
```diff
@@ -0,0 +1,164 @@
+# This file is automatically generated from KLR
+# Manual edits will be overwritten
+
+from typing import NamedTuple, Optional
+from enum import Enum, auto
+
+class Pos(NamedTuple):
```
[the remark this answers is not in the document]
govereau: Issues with Cython (we discussed briefly on chime). Also, I want these to be immutable.

interop/nkic/region.c
```diff
@@ -0,0 +1,77 @@
...
+  size_t size;
+  size_t offset;
+  struct block *next;
+  u8 buf[0];
+};
+
+#define BLOCK_SIZE (8192 - sizeof(struct block))
+#define LARGE_SIZE 7168
+
+struct region {
```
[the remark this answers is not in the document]
govereau: Correct.

interop/nkic/ast_nki.py
```diff
@@ -0,0 +1,164 @@
...
+  pass
+
+class Value_tensor(NamedTuple):
+  shape : list[int]
+  dtype : str
+
+Value = Value_tensor | Value_ellipsis | Value_string | Value_float | Value_int | Value_bool | Value_none
+
+class BinOp(Enum):
+  Land = auto()
```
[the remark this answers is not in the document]
govereau: Yes, I agree. First I need to implement the tags in Lean, then we can generate them here and in the C code.

interop/nkic/region.c
```diff
@@ -0,0 +1,77 @@
...
+  if (region) {
+    free_blocks(region->blocks);
+    free_blocks(region->large);
+    free(region);
+  }
+}
+
+void *region_alloc(struct region *region, size_t size) {
+  if (unlikely(!region))
+    return NULL;
```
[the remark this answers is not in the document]
govereau: I think we have to be careful as we are running in the user's environment. If we assert, I guess PyTorch (Jax) will crash?

interop/nkic/frontend.c
```diff
@@ -0,0 +1,146 @@
...
+static int kernel_init(struct kernel *self, PyObject *args, PyObject *kwds) {
+  (void)kwds;
+  PyObject *f = NULL;
+  if (!PyArg_ParseTuple(args, "O", &f)) {
+    // Exception set by ParseTuple
+    return -1;
+  }
+  if (!PyFunction_Check(f)) {
+    Py_INCREF(PyExc_TypeError);
+    PyErr_SetString(PyExc_TypeError, "parameter must be a function");
```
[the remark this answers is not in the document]
govereau: I have seen both ways. Looks like it doesn't matter. I will remove these increments.

interop/nkic/region.c
```diff
@@ -0,0 +1,77 @@
...
+
+  // check for large block
+  if (unlikely(size > LARGE_SIZE)) {
+    struct block *b = alloc_block(size);
+    if (unlikely(!b))
+      return NULL;
```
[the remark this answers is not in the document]
govereau: I agree in principle. We should check what happens to PyTorch and Jax when a loaded python extension aborts.

interop/nkic/region.c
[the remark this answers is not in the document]
graebm: Yup. It would crash. It's a required arg. If this were C++, it would be passed by `&` and we wouldn't need to stress about this. But it's C, so we have to pass by `*`. Aborting if another C function calls this wrong seems like the fastest way to reveal any programmer accidents. This is related to the conversation about [handling Out Of Memory](https://github.com/leanprover/KLR/pull/99#discussion_r2061548595)...

I've worked in C codebases where every function returned an error code, because that seemed like the "correct" way to build an API: What if someone passes a required arg as NULL? Sure this basic setter function can't possibly fail now, but what if there's some theoretical future where it could?

It makes the codebase extremely bloated with error handling that can't possibly happen, so the error-handling branches are never tested, and so test-coverage metrics are always low because it's impossible to get them anywhere near 100, and so people don't notice when they're not testing error-handling branches that *can* actually happen, and so those branch have bugs that aren't caught.

And old-timers write code like "well I don't need to test this return code because I know it can never actually fail", but people new to the team don't know which functions can fail or not in real life, so they ignore return codes they shouldn't, or write code in a convoluted way to deal with the possibility of things that are impossible.

Anyway, I've said my piece. Let's discuss more in person. We might need a larger error handling strategy for this codebase anyway. Like, do we need error codes? strings? etc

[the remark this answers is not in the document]
govereau: All good points. I don't disagree, but we are way up-leveled from the concern raised here. The strategy is the following:
1. Write the code in Lean and prove that it is correct. This means all corner cases are guaranteed to be handled, and there are no worries about test coverage, uncaught errors, or algorithmic mistakes. Far, far beyond that, we will have proved that the compiler preserves the semantics of the original user program.
2. Extract the Lean code to C/C++. If the extraction is correct, then the C code enjoys the benefits of the formal proofs.
3. Build a differential random testing framework for comparing the two implementations. This gives us extra assurance and also provides testing for any hand-written code in the C code base, as we improve and extraction methods.

Long term, we are not writing any C/C++ code. We are writing Lean code, so our approach doesn't have any of the drawbacks you mention.

The current thinking on a general error handling strategy can be seen in `KLR/Core/Basic`; a simple example is the `TensorName` structure. Here we require the well-formedness conditions as part of the type construction, so there can never be an invalid `TensorName`. Then for a simple function like `AccessBasic.shape` we write the code monadically to avoid dependent matching in the running code. However, we can prove a theorem like `AccessBasic.shape.noFail` which says that the error path is unreachable. This is where we put the assert in the C code: the proved unreachable path. So the error return is whatever the `Err` monad says it is, but it _literally_ never happens.

This is our current thinking. It may change as we prove more theorems and get a feel for what we really need.

PR #100, opened 2025-04-24
feat: Add lean bindings for gzip

seanmcl:
Return a ByteArray with the gzip header. (NEFF files are a 1024 bit NEFF header followed by a gz file.)


lakefile.lean
```diff
@@ -3,17 +3,22 @@ open Lake DSL
...
 
+@[default_target]
 lean_lib "KLR" where
   defaultFacets := #[LeanLib.staticFacet]
 
 @[default_target]
 lean_exe "klr" where
   nativeFacets := fun _ => #[Module.oFacet]
   root := `Main
+  moreLinkArgs := #["-L/opt/homebrew/opt/zlib/lib", "-lz"]
```
seanmcl, line 14: Going to see if Lake devs know a good way of finding shared libs.

KLR/Util/Gzip/Main.lean
```diff
@@ -0,0 +1,71 @@
+import Cli
+import Gzip
+
+open Cli
+
+private def bytesToHex (hash : ByteArray) : String := Id.run do
```
seanmcl, line 6: Hoping to migrate these bit-twiddling functions to Util at some point soon. Not sure how to do it with multiple lakefiles yet. Probably will need another Util dir with its own lakefile we can `require`.

KLR/Util/Gzip/lakefile.lean
```diff
@@ -0,0 +1,27 @@
+import Lake
+open Lake DSL System
+
+package Gzip where
+
+@[default_target]
+lean_lib Gzip where
+  moreLinkArgs := #["-L/opt/homebrew/opt/zlib/lib", "-lz"]
```
govereau: I guess this doesn't work on Linux?

Buried in the Lake sources, I found you can say:
```lean
meta if System.Platform.isWindows then
lean_lib ...
else meta if System.Platform.isOSX then
lean_lib ...
else
lean_lib ...
```
Or I guess you could write a function to return the link directory that looks at `Platform.isOSX`, etc.

KLR/Util/Gzip/lakefile.lean
```diff
@@ -0,0 +1,27 @@
...
+  root := `Main
+
+target lean_gzip.o pkg : FilePath := do
+  let oFile := pkg.buildDir / "lean_gzip.o"
+  let srcJob ← inputTextFile <| pkg.dir / "lean_gzip.c"
+  let weakArgs := #["-I", (← getLeanIncludeDir).toString, "-I", "/opt/homebrew/opt/zlib/include"] -- TODO: better way to find neuron path
+  buildO oFile srcJob weakArgs #["-std=c11", "-fPIC", "-Werror"] "cc" getLeanTrace
+
+extern_lib libleannrt pkg := do
```
govereau, line 28: Is `libleannrt` a copy-paste error?

KLR/Util/Gzip/lean_gzip.c
```diff
@@ -0,0 +1,120 @@
...
+  }
+  return lean_io_result_mk_error(lean_mk_io_user_error(lean_msg));
+}
+
+LEAN_EXPORT lean_object* lean_gzip(b_lean_obj_arg input) {
+  uint8_t *c_input = lean_sarray_cptr(input);
+  size_t input_len = lean_sarray_size(input);
+  
+  // Create a pipe
+  int pipefd[2];
```
govereau, line 37: Interesting, not what I thought I would see here, but very clever.
The pipe lets you use the higher-level API easily, and you don't have to go down a level and work with blocks of memory.
One question: can't the pipe get full? Do you need to interleave writing to and reading from the pipe if the input is very large?

govereau (approved):

KLR/Util/Gzip/lakefile.lean
seanmcl: Ah we found the same solution

seanmcl:
Finally got the ubuntu stack thing working. Will sync with FRO office hours to see how to turn protection back on, but for now it's building and running fine.

PR #101, opened 2025-04-24
feat: add CPython parser to nkic

govereau:
This change adds a private version of the CPython PEG parser to nkic: the C version of the KLR front-end. The parser is used by the gather step, which will be completed in a follow-up commit.

Note to reviewers: only the files in `interop/nkic` need to be reviewed. The files in `interop/nkic/peg_parser` are from CPython.


interop/nkic/frontend.c
```diff
@@ -34,12 +30,23 @@ static int kernel_init(struct kernel *self, PyObject *args, PyObject *kwds) {
...
+  }
   return 0;
 }
 
 // Custom deallocator for Kernel type
 static void kernel_dealloc(struct kernel *self) {
+  if (!self) return;
   Py_XDECREF(self->f); // NULL is OK
+  if (self->region)
+    region_destroy(self->region);
```
seanmcl, line 49: nit: could we use braces here?

seanmcl (approved):

interop/nkic/frontend.c
govereau: Yeah, I guess that is better style. I will start doing that.

interop/nkic/gather.c
```diff
@@ -0,0 +1,85 @@
...
+      !line || !PyLong_Check(line) ||
+      !src  || !PyUnicode_Check(src))
+    goto done;
+
+  long lineno = PyLong_AsLong(line);
+
+  Py_ssize_t size = 0;
+  const char *s = PyUnicode_AsUTF8AndSize(src, &size);
+  if (!s || size <= 0)
+    goto done;
```
graebm, line 66: if `size <= 0`, NULL will be returned, but no exception will have been raised

interop/nkic/gather.c
```diff
@@ -0,0 +1,85 @@
...
+        f = PyObject_GetAttr(m, attr);
+    }
+  }
+  Py_XDECREF(fe);
+  Py_XDECREF(m);
+  Py_XDECREF(attr);
+  return f;
+}
+
+static struct _mod* parse_function(PyObject *f) {
```
graebm, line 31: Just because we're make a Python C extension, doesn't mean we always have to use C. This functions shows what a pain it is. 35 lines of C code, to do 1 line of python:
```
file, line, src = _get_src(self._f)
```

Plus it required a 15 more lines of C code for the `get_util()` helper function. Plus there are bugs like memory leaks, and failure to raise exceptions in certain branches.

If the public API for `frontend` was all Python code, it would be easier to sanitize the input before passing it down to C.

We could have `frontend.py` or `frontend/__init__.py` for the public API. And put the C code in a private module named like `_frontend`.

So like `Kernel` could just be a python class. Its `__init__()` would be like:
```
class Kernel:
    def __init__(self, f: Callable):
        """
        Hey look it's easy to do docstrings and type hints
        """

        # sanitize input using nice easy Python code
        if not callable(f):
            raise TypeError("parameter must be a function")

        filename, line, src = _get_src(f)
        if len(src) == 0:
            raise ValueError("function must not be empty")

        # C stuff is in private module, so only our own code should be calling it directly
        _frontend.try_parse_string(src, filename)

        self._f = f
```

if we need to create pure C structs, and have them live within the python `Kernel`, we can use [PyCapsule](https://docs.python.org/3/c-api/capsule.html):
```
        # Call into C to long-lived C stuff, which gets returned wrapped in PyCapsule.
        self._capsule = _frontend.new_kernel_capsule()
```

interop/nkic/gather.c
```diff
@@ -0,0 +1,85 @@
...
+  if (!getsrc)
+    return NULL;
+
+  PyObject *args = PyTuple_New(1);
+  if (!args) {
+    Py_DECREF(getsrc);
+    return NULL;
+  }
+
+  PyTuple_SetItem(args, 0, f);
```
graebm, line 44: 1) important: [PyTuple_SetItem()](https://docs.python.org/3/c-api/tuple.html#c.PyTuple_SetItem) "steals" a reference.
2) trivial: use [PyTuple_SET_ITEM()](https://docs.python.org/3/c-api/tuple.html#c.PyTuple_SET_ITEM). You're right that `PyTuple_SetItem()` can't fail in this case, so it's not worth checking the return code, and for that exact reason there's another version of this function that doesn't have a return code (but asserts on misuse)

```suggestion
  Py_INCREF(f);
  PyTuple_SET_ITEM(args, 0, f); /* steals reference to item */
```

interop/nkic/gather.c
graebm: My last project did it that way, here are some docs I wrote about our "bindings design":
https://github.com/awslabs/aws-crt-python/blob/main/guides/dev/writing-code.md#bindings-design

seanmcl: I think part of the point of this work is to make it completely independent of the caller's python environment. That is indeed imo worth 35 lines of code.

govereau: A reasonable point. In fact this is how we started, but we were calling directly into Lean and not C. We changed this  here: https://github.com/leanprover/KLR/commit/808eff25c202bbea09629c18f89a7c605d12cd8d, and eliminated the extension module completely, in favor of a simple process IPC.

You can see a pure python version of gather in the current repo here: `interop/klr/kernel.py`. This is how the Lean version of KLR works, but we are moving away from this strategy. One reason is what Sean mentioned, the syntax of NKI depends on which version of python you happen to have---an untenable situation if you are trying to define a language.

Also, we plan to change the syntax of NKI in a way that the standard parser (through the `ast` module) can't handle, hence the vendoring of the CPython parser, and the (re)introduction of the extension module. The next step for the parser will be C code entirely extracted from a Lean specification of the grammar. This code will not rely on Python at all. Basically, you will just give the name of the file and the name of the function and it will handle everything. This version will also be usable as a stand-alone command-line tool (similar to the Lean version of KLR), that does not depend on python (or libpython) in any way. In this next step, the extension module will be a single function:

```python
out_file = _frontend.run_klr(f.__code__.co_filename, f.__name__)
```

The plan for Penguin is also to rewrite the back-end of the compiler in C++. Once that is done, we will really have zero python in the implementation of NKI: we don't need to read the serialized form in Python as only C++ code will need to do this. I agree, the current state is a bit funny, but it is an intermediate step along the way.

interop/nkic/gather.c
govereau: Good catch, I changed this.

We should be careful about relying on assert though. These are disabled in production builds of python.

interop/nkic/gather.c
graebm: I get that we're moving the guts of all this down into the C/C++ extension, so the user's Python's environment isn't involved with the actual compilation.

I was just proposing we hoist the public APIs of *this* module up into Python, so we can do simple stuff like cleaning up arguments, before passing the critical bits down into C/C++, so that C/C++ can minimize how much it needs to call back into Python.

It's not a huge deal. We can always refactor this later, if it seems like it would help

PR #104, opened 2025-04-25
feat: implement simplify pass in Lean (WIP)

govereau:


KLR/NKI/Simplify.lean
```diff
@@ -0,0 +1,275 @@
...
+
+instance : MonadExcept String Simplify where
+  throw msg := throw (.raw msg)
+  tryCatch m f := tryCatchThe SimplifyError m (fun e => f e.msg)
+
+private def withPos (pos : Pos) (m : Simplify a) : Simplify a :=
+  let decorate : SimplifyError -> SimplifyError
+  | .raw msg => .located pos msg
+  | .located p m => .located p m
+  EStateM.adaptExcept decorate m
```
seanmcl, line 43: ooh nice

KLR/NKI/Simplify.lean
```diff
@@ -0,0 +1,275 @@
...
+  | .lor => .lor
+
+private def cmpOp : Python.CmpOp -> Simplify BinOp
+  | .eq => return .eq
+  | .ne => return .ne
+  | .lt => return .lt
+  | .le => return .le
+  | .gt => return .gt
+  | .ge => return .ge
+  | .is | .isNot => throw "the is operator is not supported in NKI, use =="
```
seanmcl, line 82: Why is `a is b` allowed in the nki input, but `lambda' is not? I'd think both would be allowed or both rejected.

KLR/NKI/Simplify.lean
```diff
@@ -0,0 +1,275 @@
...
+  | .invert => return call "invert"
+  | .not => return call "not"
+  | .uadd => return fun e => e.expr
+  | .usub => return call "negate"
+
+private def binOp : Python.BinOp -> Simplify BinOp
+  | .add => return .add
+  | .sub => return .sub
+  | .mul => return .mul
+  | .matmul => throw "the matmul operator is not supported in NKI"
```
seanmcl, line 99: What is this? I see it in Python.BinOp but don't know what it corresponds to in Python.

KLR/NKI/Simplify.lean
```diff
@@ -0,0 +1,275 @@
...
+
+def SimplifyError.msg : SimplifyError -> String
+  | .raw msg | .located _ msg => msg
+
+instance : ToString SimplifyError where
+  toString
+    | .raw m => m
+    | .located p m => s!"{repr p}:{m}"
+
+abbrev Simplify := EStateM SimplifyError Unit
```
seanmcl, line 28: Why do you want a state monad with Unit as the state?

KLR/NKI/Simplify.lean
```diff
@@ -0,0 +1,275 @@
...
+  | .floor => return .floor
+
+private def booleanOp (op : BinOp) : List Expr -> Simplify Expr
+  | [e] => return e
+  | e :: es => return ⟨ .binOp op e (<- booleanOp op es), e.pos ⟩
+  | _ => throw "invalid boolean expression"
+
+private def compare : Expr -> List BinOp -> List Expr -> Simplify Expr
+  | l, [op], [y] => return ⟨ .binOp op l y, l.pos ⟩
+  | l, op :: ops, e :: es => return ⟨ .binOp op l (<- compare e ops es), l.pos ⟩
```
seanmcl, line 117: I see, so we're still allowing chains of comparisons in NKI, which are linearized in KLR. This seems like one thing you could simplify further, but I get it's a spectrum.

KLR/NKI/Simplify.lean
```diff
@@ -0,0 +1,275 @@
...
+  match e' with
+  | .const c => return .value (value c)
+  | .tensor s dtype => return .value (.tensor (<- shape s) dtype)
+  | .name s _ => return .var s
+  | .attr e id _ => do
+      match <- expr e with
+      | ⟨ .var s, _ ⟩ => return .var (s ++ "." ++ id)
+      | e => return .proj e id
+  | .tuple l _
+  | .list l _ => return .tuple (<- exprs l)
```
seanmcl, line 163: yay, list is gone!

seanmcl (commented):
Neat!

KLR/NKI/Simplify.lean
govereau: The expression `a is b` is pointer equality, I wanted to only allow structural equality in NKI.
I am not seeing the connection to lambda. Lambda is filtered out earlier because it is its own expression form.

KLR/NKI/Simplify.lean
govereau: Took me a while to figure this out, but you can write `a @ b` in python and you get the matmul operator. Numpy supports this, and I don't see why we shouldn't. However the current NKI disallows the matmul operator. I think the reason is historical: the original `nisa.matmul` didn't return a result, but took a `dst` argument, which doesn't really fit with the operator notation.

KLR/NKI/Simplify.lean
govereau: Good call. I thought I might need some state, but it looks like simplify doesn't need any. I can convert this to a simple `Except` monad.

KLR/NKI/Simplify.lean
govereau: I considered converting it down to `ifExp` to make the short-circuit explicit. Maybe that is a good idea...

KLR/NKI/Simplify.lean
```diff
@@ -0,0 +1,450 @@
...
+
+abbrev PosM := EStateM PosError PosState
+
+namespace PosM
+
+instance : MonadExceptOf String PosM where
+  throw msg := throw (.raw msg)
+  tryCatch m f := tryCatch m (fun e => f e.msg)
+
+def withPos (pos : Pos) (m : PosM a) : PosM a :=
```
seanmcl, line 114: I don't understand the Lean postfix rules. I assumed that the PosM argument had to be first for `m.withPos p` to work but evidently that is not the case. Can it just appear anywhere in the argument list?

KLR/NKI/Simplify.lean
seanmcl: 2 way door. Can leave it for now.

seanmcl (approved):

KLR/NKI/Simplify.lean
```diff
@@ -0,0 +1,450 @@
...
+  deriving Inhabited, Repr
+
+namespace PosError
+
+def msg : PosError -> String
+  | .raw msg | .located _ msg | .absolute _ _ msg => msg
+
+def locate (pos : Pos) : PosError -> PosError
+  | .raw s => .located pos s
+  | err => err
```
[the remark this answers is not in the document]
govereau: Right, and locate is not supposed to change any position already in place. I will add the other cases explicitly though with some comments.

KLR/NKI/Simplify.lean
```diff
@@ -0,0 +1,450 @@
...
+
+def addFile (file : String) (offset : Nat) : PosError -> PosError
```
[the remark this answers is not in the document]
govereau: Yup, will do.

KLR/NKI/Simplify.lean
```diff
@@ -0,0 +1,450 @@
...
+namespace PosState
+
+def warn (msg : String) (ps : PosState) : PosState :=
+  { ps with
+    newWarns := ps.newWarns.push (.raw msg)
+  }
+
+def locate (pos : Pos) (ps : PosState) : PosState :=
+  { ps with
+    warnings := ps.warnings.append (ps.newWarns.map (.locate pos))
```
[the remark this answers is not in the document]
govereau: That is what I have in mind, I will add more notes so it is clear. For instance, we should always have at least one `addFile` or we can lose warnings.

KLR/NKI/Simplify.lean
govereau: Turns out, Lean is awesome. Postfix will use the first argument that has the right type.

PR #105, opened 2025-04-26
feat: Add simple libarchive (aka "tar") interface to Lean

seanmcl:
Note how this creates/extracts in `ByteArray`s so we can take a Lean data structure representing the NEFF, convert the files to `ByteArray`s, create the archive, gzip it, plop the NEFF header on and write to disk.

### DIFF
--- a/KLR/Util/Archive/Main.lean
+++ b/KLR/Util/Archive/Main.lean
@@ -1,0 +1,89 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Paul Govereau, Sean McLaughlin
+-/
+
+import Archive
+import Cli
+
+open Cli
+open System(FilePath)
+open Util
+
+def create (p : Parsed) : IO UInt32 := do
+  let files := p.variableArgsAs! String
+  let outfile := p.flag! "output" |>.as! String
+  let mut entries := []
+  for file in files do
+    if <- FilePath.pathExists file then
+      let contents <- IO.FS.readBinFile file
+      let entry := ArchiveEntry.mk file contents
+      entries := entry :: entries
+    else
+      IO.eprintln s!"File doesn't exist: {file}"
+      return 1
+  let tarBytes := createTar entries
+  IO.FS.writeBinFile outfile tarBytes
+  IO.println s!"Wrote tar archive to {outfile} ({tarBytes.size} bytes)"
+  return 0
+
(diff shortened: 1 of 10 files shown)


KLR/Util/Archive/Main.lean
```diff
@@ -0,0 +1,89 @@
...
+  let mut entries := []
+  for file in files do
+    if <- FilePath.pathExists file then
+      let contents <- IO.FS.readBinFile file
+      let entry := ArchiveEntry.mk file contents
+      entries := entry :: entries
+    else
+      IO.eprintln s!"File doesn't exist: {file}"
+      return 1
+  let tarBytes := createTar entries
```
govereau, line 26: Does it matter that the files are in reverse order?

KLR/Util/Archive/Main.lean
```diff
@@ -0,0 +1,89 @@
...
+  IO.println s!"Extracted {entries.length} files from archive:"
+  let dir <-
+    if p.hasFlag "dir" then
+      pure $ FilePath.mk (p.flag! "dir" |>.as! String)
+    else
+      cwd
+  for entry in entries do
+    IO.println s!"File: {entry.filename}"
+    let path := System.FilePath.join dir entry.filename
+    let () <- match System.FilePath.parent path with
```
govereau, line 46: I don't think you need to say `let () <-` here.

KLR/Util/Archive/lakefile.lean
```diff
@@ -0,0 +1,62 @@
...
+
+import Lake
+open Lake DSL
+open System(FilePath)
+
+package Archive where
+
+def noStackProtector := "-fno-stack-protector"
+
+def moreLinkArgs :=
```
govereau, line 15: Nice.

KLR/Util/Archive/lean_archive.c
```diff
@@ -0,0 +1,193 @@
...
+  if (bytes < 0) {
+    fprintf(stderr, "couldn't format error message: %s", fmt);
+    lean_msg = lean_mk_string(fmt);
+  } else {
+    lean_msg = lean_mk_string(msg); // copies
+    free(msg);
+  }
+  return lean_io_result_mk_error(lean_mk_io_user_error(lean_msg));
+}
+
```
govereau, line 31: I like this, very nice. I think this is the right level of abstraction between the Lean and C.

govereau (approved):
Looks great.

KLR/Util/Archive/Main.lean
seanmcl: Probably not. Seems nice for tidiness sake.

PR #108, opened 2025-04-30
feat: ToCBOR instances for basic types

govereau:
This change adds instances for ToCBOR for basic types.

### DIFF
--- a/KLR/Serde/Basic.lean
+++ b/KLR/Serde/Basic.lean
@@ -1,0 +1,240 @@
+/-
+Copyright (c) 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Paul Govereau, Sean McLaughlin
+-/
+import KLR.Util
+
+/-
+# Encoding and Decoding of Basic CBOR values
+
+https://cbor.io
+https://www.rfc-editor.org/rfc/rfc8949.html
+-/
+
+namespace KLR.Serde
+
+class ToCBOR (a : Type u) where
+  toCBOR : a -> ByteArray
+
+export ToCBOR (toCBOR)
+
+class FromCBOR (a : Type u) where
+  parseCBOR' (arr : ByteArray) : Err (Nat × a)
+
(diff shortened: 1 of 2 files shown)


KLR/Serde/Basic.lean
```diff
@@ -0,0 +1,231 @@
...
+https://cbor.io
+https://www.rfc-editor.org/rfc/rfc8949.html
+-/
+
+namespace KLR.Serde
+
+class ToCBOR (a :Type u) where
+  toCBOR : a -> ByteArray
+
+export ToCBOR (toCBOR)
```
seanmcl, line 20: I need to make better use of `export`.

KLR/Serde/Basic.lean
```diff
@@ -0,0 +1,231 @@
...
+/-
+# Encoding and Decoding of Basic CBOR values
+
+https://cbor.io
+https://www.rfc-editor.org/rfc/rfc8949.html
+-/
+
+namespace KLR.Serde
+
+class ToCBOR (a :Type u) where
```
seanmcl, line 17: nit: space after `:`

KLR/Serde/Basic.lean
```diff
@@ -0,0 +1,231 @@
...
+  toCBOR : a -> ByteArray
+
+export ToCBOR (toCBOR)
+
+class FromCBOR (a : Type u) where
+  parseCBOR' (arr : ByteArray) : Err (Nat × a)
+
+  parseCBOR (arr : ByteArray) : Err (ByteArray × a) :=
```
seanmcl, line 25: Why are these in the class, rather than as separate definitions with instance arguments? Do you expect people will override the defaults? I only see normal uses here. IMO adding util stuff to the class makes it harder to grok.

seanmcl (approved):
Nice. I didn't read the spec, but the code looks good.

I think a round trip checker (from/to = id, ideally in both directions) for arbitrary types would be a nice test addition, but can come later.

KLR/Serde/Basic.lean
govereau: Good call: these don't need to be in the class.

PR #109, opened 2025-04-30
feat: Add json utilities

seanmcl:
- HashMap instances for BEq, FromJson, ToJson
- Drop null values from objects for equality
- Check round tripping a data structure is the identity (mod null values)
- Encoding Nats as Json Strings

### DIFF
--- a/KLR/Util.lean
+++ b/KLR/Util.lean
@@ -6,9 +6,11 @@ Authors: Paul Govereau, Sean McLaughlin
 
 import KLR.Util.Base64
 import KLR.Util.Gzip
+import KLR.Util.Json
 import KLR.Util.SHA256
 
 namespace KLR
+
 
 /-
 The default choice for an error monad is `Except String`, used for simple
@@ -37,3 +39,6 @@ def get! [Inhabited a] (x : Err a) : a := match x with
 | .ok x => x
 
 def natDivCeil (num denom : Nat) : Nat := (num + denom - 1) / denom
+
+
+end KLR
--- a/KLR/Util/Json.lean
+++ b/KLR/Util/Json.lean
@@ -1,0 +1,88 @@
+/-
+Copyright (c) 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Paul Govereau, Sean McLaughlin
+-/
+import Lean
+import Std
+
+namespace KLR.Json
+
+open Lean(FromJson fromJson? Json ToJson toJson)
+open Std(HashMap)
+
+partial def removeNullValues : Json -> Json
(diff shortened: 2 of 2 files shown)


KLR/Util.lean
```diff
@@ -37,3 +41,76 @@ def get! [Inhabited a] (x : Err a) : a := match x with
 | .ok x => x
 
 def natDivCeil (num denom : Nat) : Nat := (num + denom - 1) / denom
+
+partial def _root_.Lean.Json.removeNullValues : Json -> Json
+| .arr a => .arr (a.map removeNullValues)
```
govereau, line 46: I guess we don't find arrays with nulls as immediate children; just worried about the objects?

KLR/Util.lean
```diff
@@ -7,9 +7,13 @@ Authors: Paul Govereau, Sean McLaughlin
 import KLR.Util.Base64
 import KLR.Util.Gzip
 import KLR.Util.SHA256
-
+import Lean
+import Std
 namespace KLR
 
+open Lean(FromJson fromJson? Json ToJson toJson)
```
govereau, line 14: Should we have `KLR.Json` namespace?

KLR/Util.lean
```diff
@@ -37,3 +41,76 @@ def get! [Inhabited a] (x : Err a) : a := match x with
...
+    node.foldM (fun m k v => do
+      let v <- fromJson? v
+      return m.insert k v
+    ) HashMap.empty
+
+instance HashMapToJson [ToJson a] : ToJson (HashMap String a) where
+  toJson m := Json.mkObj $ m.toList.map fun (k, v) => (k, toJson v)
+
+instance HashMapBeq [BEq a] : BEq (HashMap String a) where
+  beq m1 m2 := m1.toList == m2.toList
```
govereau, line 84: I am worried that m1 and m2 could be the same, but produce lists in different orders.

govereau (approved):

KLR/Util.lean
seanmcl: Yes.

KLR/Util.lean
seanmcl: Yes, bug.

KLR/Util.lean
seanmcl: Yeah, I tried to document with the name. I meant "value" as in "object value, as opposed to key". Clearly I didn't do a good job. 

The problem I'm trying to solve is comparing jsons that may or may not have `"k": null` entries. It may be that all NEFF files systematically either have it or don't, but the json derivings seem to add the nulls for options. Without manually doing the codecs I'm not sure how to compare. My gut feeling is that if we are doing different logic on a missing key vs that key with a null value we are doing something wrong. 

I'm not as confident for lists, and I haven't had any problems so far so will punt.

PR #110, opened 2025-04-30
feat: Add ByteArray utils

seanmcl:
There's a lot of this in TensorLib as well. For now, let's just but byte-hacking stuff here, and think of merging it with the TensorLib versions in the future.


seanmcl:
Hmm, since we already import TensorLib, I think I'm going to try to reexport those instead.

PR #114, opened 2025-05-02
chore: fix namespaces in Util

seanmcl:
I got mixed up with modules and namespaces, so the current imports aren't what we want. For example, you'd need to

    import KLR.Util.Gzip

    def foo := Util.gzip

This change makes the namespaces more sensible.

I also left `Archive` out of the main build. It seems OK, both mac and linux should have -larchive and -lz.


seanmcl:
NRT is a different story. Still trying to figure out how to conditionally build it without requiring an extra `lake build` step in the NRT repo.

govereau (approved):

PR #124, opened 2025-05-08
feat: ListJson - control the order of json object keys

seanmcl:
Required when the Json consumer has bugs.

### DIFF
--- a/KLR/Util/ListJson.lean
+++ b/KLR/Util/ListJson.lean
@@ -1,0 +1,187 @@
+import Lean
+import Lean.Data.Format
+
+namespace KLR.Util
+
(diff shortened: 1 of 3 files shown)


KLR/Util/ListJson.lean
```diff
@@ -0,0 +1,187 @@
...
+      false
+
+def escape (s : String) (acc : String := "") : String :=
+  -- If we don't have any characters that need to be escaped we can just append right away.
+  if needEscape s then
+    s.foldl escapeAux acc
+  else
+    acc ++ s
+
+def renderString (s : String) (acc : String := "") : String :=
```
govereau, line 120: Curious why not just call `Lean.Json.renderString`?

govereau (approved):

PR #128, opened 2025-05-27
Add attributes for tile_size.* , implement Python floor div

aqjune-aws:
This patch adds a few more features that are necessary to trace `nki_matmul_tiled_`. I guess each item is worth a small thread of discussion.

1. Add tile_size.gemm_stationary_fmax and tile_size.gemm_moving_fmax: In fact I could not find their concrete values from online. (https://awsdocs-neuron.readthedocs-hosted.com/en/latest/general/nki/api/generated/nki.language.tile_size.html did not have the numbers)
Instead. I picked the constants that https://awsdocs-neuron.readthedocs-hosted.com/en/latest/general/nki/programming_model.html assume.

2. Expand Term.attr to lookup a local/global variable: This was missing from Term.attr, but I was wondering whether this was intentionally omitted because somewhere in the project I saw that this obj constructor was planned to be removed.

3. Implement Python floor div: I checked that the updated `valueOp` worked by extracting out as a scalar non-monadic function and comparing outputs on random inputs with Python.

```
def python_floordiv (l:Int) (r:Int): Int :=
  if r = 0 then 0 else
  let samesgn := (l < 0) ↔ (r < 0)
  let (l,r) := (l.natAbs, r.natAbs)
  let d := Int.ofNat (l / r)
  (if l % r = 0 then (if samesgn then id else Int.neg) d
    else (if samesgn then d else Int.neg (d + 1)))
 #eval python_floordiv 3 2       -- 1
 #eval python_floordiv 3 (-2)    -- -2
 #eval python_floordiv (-3) 2    -- -2
 #eval python_floordiv (-3) (-2) -- 1
 #eval python_floordiv 4 2       -- 2
 #eval python_floordiv 4 (-2)    -- -2
 #eval python_floordiv (-4) 2    -- -2
 #eval python_floordiv (-4) (-2) -- 2
```

### DIFF
--- a/KLR/Python.lean
+++ b/KLR/Python.lean
@@ -67,7 +67,7 @@ inductive UnaryOp where
 inductive BinOp where
   | add | sub | mul | matmul | div | mod | pow
   | lshift | rshift | or | xor | and
-  | floor
+  | floor -- the '//' operator in Python
   deriving Repr
 
 mutual
--- a/KLR/Trace/Basic.lean
+++ b/KLR/Trace/Basic.lean
@@ -122,6 +122,14 @@ private def valueOp : BinOp -> Value -> Value -> Trace Term
   | .sub, .int l, .int r => return int (l - r)
   | .mul, .int l, .int r => return int (l * r)
   | .div, .int l, .int r => return int (l / r)
+  | .floor, .int l, .int r =>
+    if r = 0 then throw "division by zero" else
+    let samesgn := (l < 0) ↔ (r < 0)
+    let (l,r) := (l.natAbs, r.natAbs)
+    let d := Int.ofNat (l / r)
+    return int (
+      if l % r = 0 then (if samesgn then id else Int.neg) d
+      else (if samesgn then d else Int.neg (d + 1)))
   | _,_,_ => throw "unimp"
 where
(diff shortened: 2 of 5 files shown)


aqjune-aws:
After this PR, in fact, `nki_matmul_tiled_` cannot be still traced because it now hits `for m in nl.affine_range(M // TILE_M):` which causes KLR to raise "unsupported loop iterator" error.

I can come up with several different approaches here:

1. Naively consider it as a sequential loop.

This will need a minimal modification on the KLR side, but if we want tracer to raise an error when Neuron just-in-time compiler raises an error, this cannot be the right behavior.

2. Run as a sequential loop when the loop body matches a specific syntactic criteria (if Neuron compiler does this).

A question here would be how faithfully we can implement the syntactic check that Neuron compiler has.

3. Consider as a sequential loop, but when the loop body is traced, attach every memory accesses with a special "thread id". In Core, if there is write-read or write-write data race between different thread ids, make it raise undefined behavior.

-

The first question is how faithfully we should reimplement the behavior of the Neuron compiler; the hardness of implementation will certainly be relevant to the answer.

govereau (approved):

PR #130, opened 2025-05-27
feta: Add Enum syntax and deriving

seanmcl:


govereau (approved):
I know this code is just being moved and already reviewed.

PR #133, opened 2025-05-28
feat: Add FromBytes class and deriving

seanmcl:


seanmcl:
Build is failing here due to some small changes to TensorLib: https://github.com/leanprover/TensorLib/pull/76

KLR/Util/FromBytes.lean
```diff
@@ -0,0 +1,101 @@
...
+end FromBytes
+
+def mkFromBytesHeader (indVal : InductiveVal) : TermElabM Header := do
+  mkHeader ``FromBytes 1 indVal
+
+def mkFromBytesBody (e : Expr): TermElabM Term := do
+  let indName := e.getAppFn.constName!
+  let env <- getEnv
+  let str := Lean.getStructureInfo env indName
+  let fields := str.fieldNames
```
govereau, line 62: Does this only work for structures? What if I try to derive this on an inductive type?

govereau: There is a `isStructure` function if you want to rule out inductive types.

seanmcl: Fixed this and added negative tests similar to NumBytes and ToBytes

PR #134, opened 2025-05-28
feat: support nl.affine_range, add nl.zeros

aqjune-aws:
This implements nl.affine_range as discussed in https://github.com/leanprover/KLR/pull/128#issuecomment-2914371722 . :) It naively implements it as a simple sequential range.

On top of it, this adds a draft of nl.zeros (https://awsdocs-neuron.readthedocs-hosted.com/en/latest/general/nki/api/generated/nki.language.zeros.html).
I wrote my initial version as adding `.const` to `Operator` in Core, that stores a constant scalar value to all elements of the destination tensor. However, it (1) gives a global change in BIR because its compilation must know it, and (2) more importantly, it was not clear whether this was allowed in the hardware ISA.
Specifically, I want to know how (2) can be checked in general - what is the documentation that I must refer to for this?

### DIFF
--- a/KLR/Extract/Extract/C.lean
+++ b/KLR/Extract/Extract/C.lean
@@ -52,6 +52,7 @@ private def genType (t : SimpleType)
   | .list .nat => "struct Nat_List*"
   | .list .int => "struct Int_List*"
   | .list .string => "struct String_List*"
+  | .list (.enum n) => s!"struct {n}_List*"
   | .list t => s!"{genType t []}_List*"
 
 private def renameVar : String -> String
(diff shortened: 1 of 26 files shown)


KLR/BIR/Compile/Operators.lean
```diff
@@ -39,6 +39,6 @@ instance : ToInst Core.TensorScalarAddr where
 def translateOperator : Operator -> Compile Inst
   | .load => return .Load { }
   | .save => return .Save { }
+  | .const => throw "translateOperator: unimplemented: .const"
```
govereau, line 42: `nl.zeros` compiles to BIR `Memset`

KLR/Eval.lean
```diff
@@ -363,6 +363,7 @@ private def evalStmt (stmt : Core.Stmt) : WithEnv Unit := match stmt with
 | .store dst Operator.save [arg] => do
   let v <- evalValue arg
   modify fun env => env.insert dst.tensor.name v
+| .store _ Operator.const args => throw s!"Unimplemented: store const {repr args}"
```
govereau, line 366: Should be `zeros` from `TensorLib`, but this is fine for now.

govereau (commented):
Looks OK, bit we should use `Memset`.

aqjune-aws:
I chose to add the constant argument to `Memset` and restricted its value to be less than 2^32 when evaluating because the C++ implementation of memset has its element value of type uint32_t.

govereau (approved):

aqjune-aws:
Force push was blocked. :( I should squash commits on my side next time & rename commit title.

PR #141, opened 2025-06-04
feat: Add a basic type system for NKI

paulmure:


KLR/NKI/Types.lean
```diff
@@ -0,0 +1,54 @@
...
+
+`nnat` is the total number of universal quantifiers in scope for `SNat`s.
+
+`ntyp` is the total number of universal quantifiers in scope for `STyp`s.
+
+Example:
+`∀ n m n, ∀ t1 t2, tensor[t1, 10, m]` ≡ `(STyp.tensor [SNat.const 10, SNat.param 1] (STyp.param 1) : STyp 3 2)`
+-/
+inductive STyp (nnat ntyp : Nat)
+  | param (idx : Fin nnat)
```
seanmcl: Not sure about encoding the number of variables with dependent types. An external type checker seems easier for me to imagine working, but I look forward to seeing more examples. For example, it would be neat to typecheck something like

```
def transpose(x):
  return np.transpose(x, (3, 1))
```

This should accept a tensor of at least 4 dimensions, and type as something like
```
all n0 n1 . tensor[n0, n1, n2, n3, ...] -> tensor[n0, n3, n2, n1, ...]
```

How would this look with the dependent types above?

govereau:
Can you add the other version with variable names? We may want this just to represent the syntax that users type.

KLR/NKI/Types.lean
paulmure: You would need to quantify over `snat` and `styp` like so:

```lean
def List.swap (l : List α) (i₁ i₂ : Fin l.length) : List α :=
  (l.set i₁ l[i₁]).set i₂ l[i₂]

def transpose {nnat ntyp : Nat} {dtype : STyp nnat ntyp} {shapes : List (SNat nnat)}
  (hshapes : shapes.length > 3) : STyp nnat ntyp :=
  have h1 : 1 < shapes.length := by omega
  have h3 : 3 < shapes.length := by omega
  .func (.tensor shapes dtype) (.tensor (shapes.swap ⟨1, h1⟩ ⟨3, h3⟩ ) dtype)
```

However, you need to prove the length of `shapes` regardless in order to index into the `list`.

seanmcl: I'd prefer to not do theorem proving when writing functions. We are trying to keep proofs and code entirely separate. This was suggested as a best practice by the CompCert folks and we've been trying to stick to that.

paulmure: Also, I should clarify that the need for less-than proofs in the transpose example has nothing to do with dependent De Bruijn indices but the need to index into lists. I could have defined it like this:
```lean
def List.swap (l : List α) (i₁ i₂ : Nat) : List α :=
  if h : i₁ < l.length ∧ i₂ < l.length then
    (l.set i₁ l[i₂]).set i₂ l[i₁]
  else
    l

def transpose {nnat ntyp : Nat} {dtype : Core.Dtype} {shapes : List (SNat nnat)} : STyp nnat ntyp :=
  .func (.tensor shapes dtype) (.tensor (List.swap shapes 1 3 ) dtype)
```

KLR/NKI/Types.lean
```diff
@@ -0,0 +1,137 @@
...
+  inductive Srt
+    | snat
+    | styp
+
+  structure TypParam where
+    name : String
+    sort : Srt
+
+  inductive SNat
+    | param (name : String)
```
seanmcl, line 71: So iiuc, the difference here is that in the dependent version we have a pointer to the implicitly bound variables, while here we just use a variable name. Seems OK to me, since at least for now we are inventing all these names on the fly; there is no concrete syntax for users to give names that may collide.

KLR/NKI/Types.lean
```diff
@@ -0,0 +1,137 @@
...
+  | func (dom ran : STyp nnat ntyp)
+
+namespace NominalTypes
+  /-!
+  # An alternative types definition using nominal variable binding
+
+  This requires a separate well-formedness proof to show that varaible references are valid.
+  -/
+
+  inductive Srt
```
seanmcl, line 62: Just checking my understanding. We are distinguishing between Styp and Snat so we can keep tensor shape constants/variables apart from regular types for expressions, and vice versa. I had not thought of having different variables for these two things but makes sense.

KLR/NKI/Types.lean
```diff
@@ -0,0 +1,137 @@
...
+def List.swap (l : List α) (i₁ i₂ : Nat) : List α :=
+  if h : i₁ < l.length ∧ i₂ < l.length then
+    (l.set i₁ l[i₂]).set i₂ l[i₁]
+  else
+    l
+
+open KLR NKI
+
+def transpose {nnat ntyp : Nat} {dtype : Core.Dtype} {shapes : List (SNat nnat)} : STyp nnat ntyp :=
+  .func (.tensor shapes dtype) (.tensor (List.swap shapes 1 3 ) dtype)
```
seanmcl, line 135: Nice example. So for sth like matmul2D we'd have this?
```
matmul2D : .func2 (.tensor [l, x] dtype) (.tensor [x, r] dtype) (.tensor [l, r] dtype)
```

seanmcl (commented):
Nice, I prefer this new one, though more examples would help. Maybe you could tal at https://github.com/aws-neuron/nki-samples/blob/main/src/nki_samples/tutorials/average_pool2d/average_pool2d_nki_kernels.py, which has a mix of challenges.

KLR/NKI/Types.lean
paulmure: This is the age-old debate over variable binder representation. Nominal representations like this has issues with alpha equivalence. De Bruijn indices have nicer mathematical representation but dealing with incrementing/decrementing the numbers can be annoying.

My idea here is that when a type is being defined, all parameters are given up front. So we don't have the issue with changing the De Bruijn indices as you would when representing something like nested lambda functions. Nominal logic in general is a pain so I'd like to avoid them if possible.

I am also not aware of a way to explicitly provide type parameters in python. I don't think things like this is possible:
```python
x : tuple[a, b] = (1, 'str')
```
You can do this:
```python
x : tuple[any, any] = (1, 'str')
```
I'd argue that this is not actually a type parameter but a special `any` type.

All this means that type parameters will likely just be inferred, which I think further advocates for the use of De Bruijn indicies (which can be done without dependent types if we want).

seanmcl: Works fine?


```
from typing import Tuple
x: Tuple[int, str] = (1, 'foo')
```

KLR/NKI/Types.lean
paulmure: Btw, this is `matmul2d` in the nominal representation:
```lean
-- `∀ {s1 s2 s3} {t}, tensor[t, s1, s2] → tensor[t, s2, s3] → tensor[t, s1, s3]`
def nmatmul2d {dtype : Core.Dtype} : Typ :=
  ⟨
    [⟨"s1", .snat⟩, ⟨"s2", .snat⟩, ⟨"s3", .snat⟩, ⟨"t", .styp⟩],
    .styp <|
      .func (.tensor [.param "s1", .param "s2"] dtype)
        <| .func (.tensor [.param "s2", .param "s3"] dtype) (.tensor [.param "s1", .param "s1"] dtype)
  ⟩
```

seanmcl: What is the `t` in the comment, e.g. `tensor[t, s1, s2]`?

KLR/NKI/Types.lean
paulmure: > If we're inferring all the types, and can thus control variable names, why is debruijn notation still better? Is it just easier to do proofs when you don't need to keep saying things like "x" is not in the set of names we've seen so far?

That's my thinking. The De Bruijn version is well-formed by construction rather than by a separate proof.

paulmure: > Works fine?
> 
> ```
> from typing import Tuple
> x: Tuple[int, str] = (1, 'foo')
> ```

But that doesn't quantify over the two types inside the tuple. I don't think you can do something like this:
```python
def swap(t : tuple[a, b]) -> tuple[b, a]:
    return (t[1], t[0])
```
Instead, you can do this:
```
def swap(t : tuple[any, any]) -> tuple[any, any]:
    return (t[1], t[0])
```
Which is not real polymorphism in the since we'd like to capture in the type system.

KLR/NKI/Types.lean
seanmcl: 1. I guess we should use the Python convention of tuple-ing the arguments. Assuming one important reason for the type checker is to help NKI authors, they will be writing Python and currying is not super common imo. 
2. I'd thought about this in the context of allowing a function to take "castable" dtypes, e.g. adding Int32 and Int64 tensors without an explicit `astype`. I gave up on that when I figured we could have the compiler insert `astype` calls either at the array level or when indexing over the two tensors. But still, it's worth thinking about the pros and cons.

KLR/NKI/Types.lean
```diff
@@ -0,0 +1,397 @@
...
+  /-- Infer the De Bruijn indexed version -/
+  def STyp.inferDeBruijn (t : STyp) : Option (NKI.STyp (TypParam.nnat t.inferBinders) (TypParam.ntyp t.inferBinders)) :=
+    Typ.toDeBruijn ⟨t.inferBinders, .styp t⟩
+
+  /- Parser for type annotations (`pt_` prefix for Py Types) -/
+
+  open Lean Elab Meta
+
+  /- SNat -/
+  declare_syntax_cat pt_nat
```
seanmcl, line 269: Ooh this is awesome.

KLR/NKI/Types.lean
```diff
@@ -0,0 +1,457 @@
...
+  syntax pt_nat ", " pt_nat_tup : pt_nat_tup
+
+  declare_syntax_cat pt_typ_tup
+  syntax pt_typ : pt_typ_tup
+  syntax pt_typ ", " pt_typ_tup : pt_typ_tup
+
+  syntax "tuple[" pt_typ_tup "]" : pt_typ
+  syntax "tensor[" pt_dtype ", " "(" pt_nat_tup ")" "]" : pt_typ
+
+  syntax pt_typ " → " pt_typ : pt_typ
```
seanmcl, line 316: nit: can we allow ascii as well, ->. I have an irrational allergy to unicode.

paulmure: Done!

paulmure:
Closing this for the initial draft. More polished version will come along after we decide how general the type system should be.

PR #147, opened 2025-06-08
chore: change representation of C types

govereau:
This patch changes the C type generator so that we always use a pointer when referring to a structure. This makes the code generation more regular, and should be easier to deal with in future code generation meta-programs.


KLR/Extract/Extract/C.lean
```diff
@@ -36,51 +44,35 @@ private def CName (name : Name) (f : String -> String := id) : String :=
 
 instance : ToString Name where toString n := CName n
 
-private def genType (t : SimpleType)
-                    (indirect : List Name := [])
-                    : String :=
```
govereau, line 41: The only real change is to this function `genType`. I removed the `indirect` hack and just always use a pointer when creating a variable of struct type. Everything else is necessary changes downstream of this.

seanmcl (approved):

PR #149, opened 2025-06-09
feat: add ascii string <-> bitvec conversions

seanmcl:


KLR/Util/Common.lean
```diff
@@ -0,0 +1,43 @@
+/-
```
govereau, line 1: Oh good. This file has caused problems for me where it was before.
By importing `Util` and `KLR.Util`

govereau (approved):

PR #163, opened 2025-06-16
feat: Allow un-numbered enum fields

seanmcl:
Enum is starting to provide value, so let's not force people to add ints when they don't matter.

### DIFF
--- a/KLR/Util/Enum.lean
+++ b/KLR/Util/Enum.lean
@@ -4,15 +4,17 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Paul Govereau, Sean McLaughlin
 -/
 import Lean
+import TensorLib.Common
 import Util.Hex
 
 open KLR.Util.Hex(encode)
-open Lean(Syntax TSyntax TSyntaxArray ToJson mkIdent)
-open Lean.Elab.Command(CommandElab elabCommand liftTermElabM)
+open Lean(FromJson Json Syntax TSyntax TSyntaxArray ToJson mkIdent fromJson? toJson)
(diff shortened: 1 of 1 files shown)


KLR/Util/Enum.lean
```diff
@@ -35,18 +37,24 @@ private def u8ToHex (x : UInt8) : String := "0x" ++ (encode ⟨ #[x] ⟩).toUppe
...
 where
-  doit (priv : Bool) (name : TSyntax `ident) (items : TSyntaxArray `item) := do
+  doit (priv : Bool) (name : TSyntax `ident) (items : TSyntaxArray `item) : CommandElabM Unit := do
     let mut values : List (TSyntax `Lean.Parser.Command.ctor × TSyntax `num) := []
     let mut ctors : Array (TSyntax `Lean.Parser.Command.ctor) := #[]
+    let mut ctr : Nat := 0
     for item in items do
       match item with
-      | `(item| $id := $v) => do
+      | `(item| $id:ctor := $v:num) => do
```
govereau, line 50: I know we discussed this on one of my PRs, but just wondering why you didn't set `ctr = v+1` here?

govereau (approved):
Nice!

KLR/Util/Enum.lean
seanmcl: I'm not sure that one is better than the other? For example, if we have 
```
enum Foo where
| x
| y
| z := 0
```

it's still going to fail. I guess the right solution is to look at all the numbers that are assigned and backpatch numbers after you know which ones aren't taken. I'll do that.

seanmcl: Done. Indeed this is better.

PR #170, opened 2025-06-19
feat: Add Sexp and ToSexp deriving

seanmcl:
Following Ocaml sexp serialization: https://dev.realworldocaml.org/data-serialization.html

FromSexp deriving will be harder, so is coming in another review. I think I'd like to allow parsing structure sexps of the form

```
structure Foo where
  x : Int
  y : Bool

((x 5) (y true))

in addition to

(5 true)

Comments welcome.

### DIFF
--- a/KLR/Util.lean
+++ b/KLR/Util.lean
@@ -19,5 +19,7 @@ import Util.NumBytesTest
 import Util.Padding
 import Util.Plausible
 import Util.SHA256
+import Util.Sexp
+import Util.SexpTest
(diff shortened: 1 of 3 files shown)


seanmcl:
This doesn't handle `mutual` data structures, which could be supported fairly easily if you know what you are doing. I don't have a use case yet, and only 50% know what I am doing so am going to punt.

seanmcl:
Going to rework this a bit. Sorry for the noise.

PR #171, opened 2025-06-19
feat: add FromCBOR deriving

govereau:
Similar to ToCBOR---handles mutual inductives.

### DIFF
--- a/KLR/Serde/Basic.lean
+++ b/KLR/Serde/Basic.lean
@@ -28,6 +28,11 @@ class FromCBOR (a : Type u) where
 
 open FromCBOR (parse)
 
+-- This is used by the derived instances
+def parseCBOR' [FromCBOR a] (arr : ByteArray) (size : Nat) : Err (ByteArray × Nat × a) :=
+  (parse arr).map fun (sz, x) =>
+    (.mk (arr.data.drop sz), sz + size, x)
+
 def parseCBOR [FromCBOR a] (arr : ByteArray) : Err (ByteArray × a) :=
   (parse arr).map fun (sz, x) =>
     (.mk (arr.data.drop sz), x)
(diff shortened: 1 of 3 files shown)


KLR/Serde/Test.lean
```diff
@@ -125,26 +125,36 @@ Deriving tests
...
   | n : Nat -> Y a
   | x : X a -> Y a
-  deriving ToCBOR
+  deriving ToCBOR, FromCBOR
 end
 
+
+#check X.fromBytes
+#print X.fromBytes
+#eval X.fromBytes (ByteArray.mk #[ 1 ])
```
govereau, line 154: Error is here: "type class instance problem is stuck".

KLR/Serde/Elab.lean
```diff
@@ -107,10 +108,73 @@ def mkToInstances (names : Array Name) : TermElabM (Array Command) := do
...
+    let cases <- `(match vt with
+      $arms:matchAlt*
+      | _ => throw s!"unexpected value tag {vt}")
+
+    -- Generate local instances for all type constructors
+    -- They don't have to be named, only in scope
+    let ts <- getTypeParams name
+    let ls := names.foldrM fun n body => `(
+      let _ : FromCBOR ($(mkIdent n) $ts*) := ⟨$(fnIdent n "fromBytes")⟩
+      $body
```
govereau, line 137: I suspect this trick isn't working...

KLR/Serde/Basic.lean
```diff
@@ -28,6 +28,11 @@ class FromCBOR (a : Type u) where
 
 open FromCBOR (parse)
 
+-- This is used by the derived instances
+def parseCBOR' [FromCBOR a] (arr : ByteArray) (size : Nat) : Err (ByteArray × Nat × a) :=
```
seanmcl, line 32: I had this too. Annoying we can't make it private. Briefly wondered if we should come up with some convention like `def _parseCBOR_deriving_ ...` or something, but lean4 doesn't do it so probably fine.

KLR/Serde/Basic.lean
```diff
@@ -633,23 +638,24 @@ TODO: lift this constraint.
 
 def cborTag (typeTag valTag len : Nat) : ByteArray :=
   assert! len < 0x18
-  let len := toCBOR len.toUInt64
-  let len := adjustTag 0x80 len
-  .mk #[ 0xd9, typeTag.toUInt8, valTag.toUInt8] ++ len
+  --let len := toCBOR len.toUInt64
```
seanmcl, line 641: Can we not keep commented code?

seanmcl (approved):
Nice. You're still much better at this than I am, so no comments today!

PR #172, opened 2025-06-19
Add support for assignment to subscript (x[s] = e)

aqjune-aws:
This patch adds an assignment to subscript `x[s] = e`.
This is an important case because, when x is a location to the memory and e is `nki.language.load`, it must immediately store to the location (e.g., when s is ellipse, `t[...] = nl.load` must achieve a full simple copy).
Previously, when the RValue function is traversing e, it could have created a temporary access (Tensor.load), so this could not be achieved. This problem is addressed in this patch by making the `assign` function not call `RValue` and making `assignTerm` function probe whether the RHS was `.store`.

After this patch, the only remaining unsupported case in `nki_matmul_tiled_` is `res_psum += nl.matmul`.

### DIFF
--- a/KLR/Trace/Python.lean
+++ b/KLR/Trace/Python.lean
@@ -494,30 +494,18 @@ assignments. This is effectively a simple form of constant propagation and
 dead-code elimination for simple assignments.
 -/
 
--- Convert an expression in assignment context (an L-Value).
--- TODO: handle subscript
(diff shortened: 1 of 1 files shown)


aqjune-aws:
I should set enabling tracing of `nki_matmul_tiled_` possible as a close-term goal. :) I believe this won't be too hard within the remaining time frame.

PR #175, opened 2025-06-23
chore: add standard derived instances to Python and NKI types

govereau:
Add automatically derived instances of BEq, Repr, and conversions to and from CBOR, Json and Sexp to the Python and NKI abstract syntax trees.

### DIFF
--- a/KLR/Util/Sexp.lean
+++ b/KLR/Util/Sexp.lean
@@ -7,6 +7,7 @@ This is copied and modified from FromToJson.lean in the Lean sources. The bigges
 supporting both positional and named arguments.
 -/
 import TensorLib
+import Util.Float
 
 open Lean(Command ConstructorVal CoreM Expr Ident InductiveVal Macro Name Syntax TSyntax Term getConstInfoCtor getConstInfoInduct getEnv getStructureFieldsFlattened isInductive isStructure mkIdent quote)
 open Lean.Core(mkFreshUserName)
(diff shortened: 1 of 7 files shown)


KLR/Util/Sexp.lean
```diff
@@ -47,12 +47,21 @@ export ToSexp(toSexp)
 instance : ToSexp String where
   toSexp n := atom n
 
+instance : ToSexp Bool where
+  toSexp | true => atom "#t"
```
seanmcl, line 51: This is scheme-specific. Even lisp/elisp don't use it. I guess I'd prefer just true and false.

KLR/Util/SexpTest.lean
```diff
@@ -18,6 +18,18 @@ private def roundTrip [BEq a] [ToSexp a] [FromSexp a] (x : a) : Bool :=
   | .error _ => false
   | .ok z => x == z
 
+#guard roundTrip true
+#guard roundTrip false
+
+#guard roundTrip 1.3
```
seanmcl, line 24: Nice tests! Could you add a plausible test as well?

seanmcl (approved):

KLR/Util/SexpTest.lean
govereau: Added plausible tests to the new Float parser.

govereau:
Updated with custom parser for Float

seanmcl (approved):
Nice.

PR #189, opened 2025-07-01
Use CMake (instead of Makefile) for C extension

graebm:
- Use CMake instead of Makefile
    - We're using custom build files (vs simply passing a list of .c files to setuptools), so we can do fancy stuff like build standalone C tests in addition to the Python C extension.
    - CMake (instead of Makefile) will let us build on Windows in the future
- Use [scikit-build-core](https://github.com/scikit-build/scikit-build-core) (instead of setuptools) as the build backend, because it's the [recommended](https://packaging.python.org/en/latest/guides/tool-recommendations/#build-backends) way to package an extension that uses CMake.
- Clean up standalone C tests a bit
    - Tests don't need to link Python anymore.
    - Make list of "core" .c files that don't need Python, link all "core" .c files into tests, so tests don't need to `#include` individual .c files anymore.
    - We could make the tests nicer in the future, but I wanted to minimize diffs in this PR. For example, just have one test executable, where you pass the name of the unit test function you want to run. Vs the current state where we have multiple test executables and each runs many unit test functions every time.
    - We could also do a more formal split of the "core" C code that doesn't need Python, like put it in a common subdirectory. I'm not doing that in this PR to minimize diffs.


interop/klr/stdc.h
```diff
@@ -8,10 +8,6 @@ Authors: Paul Govereau, Sean McLaughlin
 // A simple header file to check the C version and bring C standard
 // definitions into scope.
```
govereau, line 10: Should we pick another version?

interop/nki/language/__init__.py
```diff
@@ -1,4 +1,4 @@
-# KLR implemetations of NKI langauge APIs
+# KLR implementations of NKI language APIs
```
govereau, line 1: Wow, 2/6 not great stats on that sentence. :)

interop/klr/Makefile
```diff
@@ -1,46 +0,0 @@
-# A simple Makefile for local development
```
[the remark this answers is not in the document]
govereau: Maybe rename it `Makefile.dev` or something ?

graebm:
Nope

PR #193, opened 2025-07-02
feat: trace `x = f()` and `x binop= f()` with fn keyword args

aqjune-aws:
This patch supports an alternative form of tracing `x = f()` and `x binop= f()` when this is necessary.

Specifically, this is for supporting lowering of `res_psum += nl.matmul(...)` NKI to KLR.

The Python tracer was first unfolding this to `res_psum = res_psum + nl.matmul(...)`, but this has two issues:

1. It has to expand `exprOp : BinOp -> Expr -> Expr -> Trace Term` when the second argument of exprOp is Core.Expr.call. However, in this case, it is not clear how the output of `exprOp` can be represented in Term. One possible solution is to temporarily store the output of `nl.matmul` into a temporary buffer (in SBUF or PSUM or anywhere) and use the access as the second argument of `exprOp`, but then the following issue appears.

2. I understood that `res_psum += nl.matmul(...)` is supposed to do an in-place update to `res_psum` (but without a clear reason; I looked into the s3d3_mm doc but could not immediately find a flag for this. But it is kind of weird if '+=' tries to store into a new memory region :/). Storing the output `nl.matmul` into a temporary buffer does not match what NKI does.

To avoid this problem, this patch tries to achieve the following lowerin:

```
t += matmul(a,b)  # becomes matmul(a, b, dst = t, accum = True)
t = matmul(a,b)   # becomes matmul(a, b, dst = t, accum = False)
```

This patch detects this `y (binop)= f()` pattern from KLR.Python.Stmt and tries to evaluate the alternative `f(..., dst = t(, accum = ...))` function call. It only does this when the function is built-in or non-user-Python function.

### DIFF
--- a/KLR/Trace/Python.lean
+++ b/KLR/Trace/Python.lean
@@ -641,6 +641,45 @@ partial def integer? : Option Expr -> Trace (Option Int)
   | none => return none
   | some e => expr e
 
+-- Deal with a function call, '.call f args kws'
+-- assignLhs is set when this call was the RHS of assignment,
+-- 'lhs = f(args, kws)'.
+partial def fnCall (f:Expr) (args:List Expr) (kws:List Keyword)
+                   (assignLhs:Option Term × Option BinOp): Trace Term := do
+  let expand_kwargs_from_assign (kwargs:List (String × Term))
+      : Trace (List (String × Term)) :=
+    match assignLhs with
+    | (.some dst, binop) => do
+      let accum := binop == Option.some BinOp.add
+      let accval:Term := .expr (.value (.bool accum)) .bool
+      return (kwargs.append [("dst", dst), ("accum", accval)])
+    | _ => do
+      return kwargs
+
+  match (<- expr f : Term) with
+  | .builtin n _ self => do
+      let f <- builtinFn n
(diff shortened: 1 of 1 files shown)


govereau:
Sorry, my fault. I think if you rebase onto 886d8ff it will work again?

aqjune-aws:
Umm, but I will receive the approval of this pull request as a message that it works on the latest KLR :)

govereau:
> Sharing the error message from `make` just in case it could be helpful:
> 
> ```
> serde.c:55:9: error: call to undeclared function 'open_memstream'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
>    55 |   out = open_memstream(&buf, &size);
> serde.c:98:8: error: call to undeclared function 'fmemopen'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
>    98 |   in = fmemopen((void*)buf, size, "r");
> ```

Oh no. OK I will have to investigate. Thanks.

govereau:
Probably I need to add `#define _GNU_SOURCE` to the relevant C file?

govereau:
Great. thanks for checking.

PR #196, opened 2025-07-02
Upload sdist to PyPI, and tweaks to support Brazil's Python3PBuildTool

graebm:
Brazil's Python3PBuildTool requires an [sdist](https://docs.python.org/3.10/distutils/sourcedist.html) of the python project, and it's currently using older version of setuptools (68.0.0).

Tweaks:
- Copy LICENSE file into python project dir (because it's simpler than relying on a custom script to inject it as the wheels are being built)
- Some tweaks to `pyproject.toml` to work with older setuptools:
    - require exact setuptools version used in Brazil, so we catch incompatibilities sooner
    - tweak how `license` was defined
    - had to move `tool.setuptools.ext-modules` definitions to `setup.py`
    
Also, fixed our publishing steps to also upload a source distribution (aka "sdist", aka `*.tar.gz` file) to PyPI, in addition to the `*.whl` files. The sdist comes from the `CI / build (ubuntu-latest)` run. I did that so the `klr.bin` included with the sdist would be one that works on x86_64 Linux (I've tried running it on Amazon Linux 2 and it seemed to work).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,12 +84,24 @@ jobs:
           MACOSX_DEPLOYMENT_TARGET: 99.0
           CIBW_BUILD_VERBOSITY: 1
         run: |
-          pip install cibuildwheel
+          pip install cibuildwheel build
           bin/make-wheel
       - uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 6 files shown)


govereau (approved):

seanmcl (approved):
Nice!

PR #198, opened 2025-07-03
chore: have sexp parser parse a list

seanmcl:
This is more flexible when we have a file like

    (set x 5)
    (f x x)

### DIFF
--- a/KLR/Util/Sexp.lean
+++ b/KLR/Util/Sexp.lean
@@ -43,10 +43,16 @@ def oneline : Sexp -> String
 instance : ToString Sexp where
   toString := oneline
 
-partial def fromString (s : String) : Err Sexp := do
-  let (e, cs) <- exp s.toList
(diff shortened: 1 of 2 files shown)


KLR/Util/Sexp.lean
```diff
@@ -43,10 +43,16 @@ def oneline : Sexp -> String
...
   toString := oneline
 
-partial def fromString (s : String) : Err Sexp := do
-  let (e, cs) <- exp s.toList
-  if cs.all Char.isWhitespace then return e
-  else throw s!"left-over stuff {cs.toString}"
+partial def fromString (s : String) : Err (List Sexp) := do
+  let mut cs := s.toList
+  let mut sexps := []
+  repeat do
```
govereau, line 49: repeat! Already applying new knowledge.

govereau (approved):

PR #204, opened 2025-07-09
Index expressions semantics and composition

markusdemedeiros:


govereau (approved):
This looks great. Thanks!

PR #205, opened 2025-07-09
Add initial KLR AST definition

P-bibs:
Opening this so that we have a place to discuss what the KLR AST should look like (@govereau @markusdemedeiros)

Outstanding TODOs:
* How much control flow do we want to handle initially, and how do we represent it in the AST
* How to properly make the AST generic over "values" (what I've called `Atom`s in the first draft)
* Add a small expression language for register operations
* Does the AST match the ISA close enough?


P-bibs:
Superseded by #214, which integrates these operators into KLR.Core

PR #211, opened 2025-07-10
chore: add CBOR and Sexp instances for Lean.Name

govereau:


seanmcl (approved):
Nice. Thanks for the Sexp stuff.

PR #212, opened 2025-07-11
Instruction integration

ppotapov-aws:
Pull in instructions to match nki isa

Most of the flags that had enums of 2 invariants are replaced with booleans.

Where we could collapse information from original enum + union structs, we collapsed down to
an inductive.

Right now there is no static information for inline arrays preserved. The generic list is used. This can
be replaced with a Vec with constraints on a size of a Vec if this information is more useful then cumbersome
to work around

For the operations we chose to define tensor view as
```
inductive TensorView where
  | literal (freePattern: List APPair) (offset : Nat := 0) (dtype : Dtype) (parNum : Nat)
  | reg (_ : Reg)
```
That should contain enough information reconstruct the original instructions with appropriate information

Most instructions have `src` and `dst` tensor views. These must have same parNum.

### DIFF
--- a/KLR.lean
+++ b/KLR.lean
@@ -11,3 +11,4 @@ import KLR.Python
 import KLR.Serde
 import KLR.Trace
 import KLR.Util
+import KLR.KLR
--- a/KLR/KLR.lean
+++ b/KLR/KLR.lean
@@ -1,0 +1,1 @@
+import KLR.KLR.Basic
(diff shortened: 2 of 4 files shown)


P-bibs:
Another note for @govereau: rather than overwrite the existing KLR definitions in KLR.Core, we put this new work in KLR.KLR. The new work likely supersedes KLR.Core and KLR.Core should be deleted, but I wanted to leave it for now for discussion since some of the definitions there appear to be needed for NKI compatibility

P-bibs:
Another consideration: currently, dimensions lists are simply `List Nat`, even though in practice the hardware only supports dimension lists (ranks) of up to 4. Should we encode this restriction in KLR, or allow arbitrary dimension lists and simply error when trying to compile to the ISA later?

govereau (approved):
Looks good. Thanks!

PR #214, opened 2025-07-14
chore: integrate operators into KLR.Core

govereau:
This change integrates the operators from KLR.KLR into KLR.Core in preparation for tracing from NKI.Basic. The definition of TensorRef wsa extended to include the abstract TensorSBuf (previously TensorName) types as a variant.


P-bibs (approved):
Looks good. I think just adding another variant to `TensorRef` is a good migration path for now, and it lets me start implementing HLR->KLR with a stable set of operators.
We still need to decide on control flow, but if that's not a blocker for NKI than we can put it off for now.

ppotapov-aws (approved):

KLR/Core/Indexing.lean
```diff
@@ -206,7 +206,7 @@ theorem AccessBasic.getShape_length (a : AccessBasic) :
   · rename_i heq; apply (AccessBasic.shape.noFail _ heq).elim
 
 @[simp] def Shape.freeDim (s : Shape) : Nat := s.freeDims.length
-@[simp] def TensorName.freeDim (s : TensorName) : Nat := s.shape.freeDim
+@[simp] def TensorSRam.freeDim (s : TensorSRam) : Nat := s.shape.freeDim
```
seanmcl, line 209: nit: SRam is a weird capitalization. I'd get Sram and SRAM. In one repo I worked in (forgot, was a while ago) there was a convention that no matter whether a word was an acronym or not, it was always CamelCased. I like that convention, but not wedded to it.

KLR/Core/Operators.lean
```diff
@@ -1,73 +1,86 @@
...
-not as arguments of the call node. A typical example is:
+These definitions follow the hardware ISA as close as possible.
+-/
+namespace KLR.Core
+open Lean (FromJson ToJson)
+open Serde (FromCBOR ToCBOR)
+open Util (FromSexp ToSexp)
 
-  .store t [] (.call (.operator (.tensorScalar ts)) [a,b] [])
+instance : ToJson Float32 where
```
seanmcl, line 22: I think these should either be marked local, or go in Common. It would be annoying to include a file and get a new class instance that was not obviously related to the types in the file I included.

KLR/Core/Operators.lean
```diff
@@ -118,58 +131,460 @@ def isArith : AluOp -> Bool
   | .bypass => true
   | op => not op.isBitwise
 
+def IsArith : AluOp → Prop
+| .abs | .add | .average | .bypass | .divide | .is_equal | .is_ge | .is_gt | .is_le | .is_lt | .max | .min | .mod | .mult | .not_equal | .pow | .rsqrt | .subtract => True
+| _ => False
```
seanmcl, line 136: Why is this going into Prop? Can we stick with Bool?

KLR/Core/Tensor.lean
```diff
@@ -0,0 +1,458 @@
...
+instance : FromCBOR TensorSRam := ⟨ fun _ => throw "" ⟩
+instance : FromJson TensorSRam := ⟨ fun _ => throw "" ⟩
+instance : FromSexp TensorSRam := ⟨ fun _ => throw "" ⟩
+instance : ToCBOR TensorSRam := ⟨ fun _ => default ⟩
+instance : ToSexp TensorSRam := ⟨ fun _ => default ⟩
+instance : ToJson TensorSRam := ⟨ fun _ => default ⟩
+
+namespace TensorSRam
+
+def make (name : String)
```
seanmcl, line 152: I wonder if there's a way to make `mk` private but still allow for FromX

seanmcl (approved):

KLR/Core/Operators.lean
```diff
@@ -118,58 +131,460 @@ def isArith : AluOp -> Bool
...
+inductive AccumCmd where
+  | Idle
+  | Zero
+  | Accumulate
+  | ZeroAccumulate
+  | LoadAccumulate
+  deriving BEq, FromCBOR, FromJson, FromSexp, Repr, ToCBOR, ToJson, ToSexp
+
+@[serde tag = 137]
+inductive ActivationFunc where
```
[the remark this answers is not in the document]
govereau: Yes, would love to use `enum`, but I need enum to pass the attribute down to the inductive it generates.

KLR/Core/Tensor.lean
```diff
@@ -0,0 +1,458 @@
...
+@[serde tag = 122]
+inductive ParQuadrant where
+  | par0 | par32 | par64 | par96
+  deriving BEq, FromCBOR, FromJson, FromSexp, Repr, ToCBOR, ToJson, ToSexp
+
+/-
+A structure representing the layout of a tensor in SRam. This maps very closely
+to the way that the ISA expects tensor accesses to be expressed. Specifically,
+it separates the parallel dimension from the rest of the dimensions and
+includes constraints like the fact that the parallel dimension stride is always 1.
```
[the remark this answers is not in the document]
seanmcl: OK, so when there's sth like

```
pub struct Tensor3d {
    pub start_addr:  Addr4,
    pub step_elem:   [i16;3],
    pub num_elem:    [u16;3],
}
```

this excludes the partition dimension, so this would actually be a 4d tensor computation?

KLR/Core/Operators.lean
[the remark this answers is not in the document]
govereau: I have seen example in the Lean compiler of matching on the attributes but I am not sure how it works in a deriving handler. I suppose I can also assign the attribute after-the-fact, but it would be a bit nicer if it just worked.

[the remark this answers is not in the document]
govereau: Oh, wait it is not a deriving handler. If you look for `docComment` in the Lean/Elab` directory you can find many examples of matching on the doc comment and the attributes.

KLR/Core/Tensor.lean
govereau: I added a comment about this. Looks like we have uses of both make and mk in the code, so need to think about it.

PR #215, opened 2025-07-14
chore: update extraction to generate KLIR files

govereau:


KLR/Extract/Extract/Basic.lean
```diff
@@ -16,24 +16,31 @@ open Lean Meta
...
+
+There is a special case for enum, which is just a constant where we know the
+type is an inductive where none of the constructors have any arguments. This
+property is only known after we collect the whole type, so there is a
+post-processing step which rewrites const to enum in those cases.
 -/
 
 inductive SimpleType where
-  | bool | nat | int | float | string
+  | bool | nat | int | float | string | prop
```
seanmcl, line 27: could we have a comment why we need prop?

KLR/Extract/Extract/Basic.lean
```diff
@@ -16,24 +16,31 @@ open Lean Meta
...
 -- For instance, placing them in a common, shared file.
 def isCommon : SimpleType -> Bool
-  | .bool | .nat | .int | .float | .string => true
+  | .bool | .nat | .int | .float | .string | .prop => true
   | .const .. | .enum .. => false
   | .option t | .list t => t.isCommon
+  | .pair l r => l.isCommon && r.isCommon
 
 -- For languages like C, we use, e.g. Bool_List instead of List Bool
 def name : SimpleType -> Name
```
seanmcl, line 46: Why isn't this in C.lean? We can reuse these names for multiple languages?

KLR/Extract/Extract/C.lean
```diff
@@ -193,6 +198,7 @@ def nkiAST : MetaM (List LeanType) := do
     `KLR.NKI.Expr,
     `KLR.NKI.Index,
     `KLR.NKI.Keyword,
+    --`KLR.NKI.Pattern,
```
seanmcl, line 201: No commented code please.

KLR/Extract/Extract/Serde.lean
```diff
@@ -240,3 +244,11 @@ def generateNkiH : MetaM Unit := do
 def generateNkiC : MetaM Unit := do
   IO.println <| C.headerC ["cbor.h", "serde_common.h", "serde_nki.h"]
   genC (<- C.nkiAST)
+
+def generateKlrH : MetaM Unit := do
+  IO.println <| C.headerH ["ast_common.h", "ast_klir.h"]
```
seanmcl, line 249: I keep expecting FileIO here, but remembered you take over stdout. Do you recommend this for other codegen projects?

seanmcl (approved):
Holy cow. 4k lines of C!

KLR/Extract/Extract/Serde.lean
govereau: I am not sure. I am starting to think what I should do is write a small meta function to build a lean value from the collected types. Then just write normal Lean code that uses the generated constant at runtime. It seems nicer to have a real program you can run rather then generating files while the lean compiler is running.

PR #217, opened 2025-07-14
chore: Have FromSexp throw on unknown fields

### LINKED ISSUE #199: FromSexp doesn't fail on unknown names
If you have a typo in your structure field that has a default, you will quietly get the default without an error.

seanmcl:
I've had several bugs with typos in names. This is especially problematic when you rename a field in Lean, but forget to update the name in text files. This change throws on unknown names.

It also introduces a small Meta library that can be a home for common patterns of metaprogramming we use, so we don't need to remember dozens of library names.

Closes #199

### DIFF
--- a/KLR/Util/Sexp.lean
+++ b/KLR/Util/Sexp.lean
@@ -10,6 +10,7 @@ import TensorLib
 import Util.Common
 import Util.Float
 import Util.Hex
+import Util.Meta
 
 open Lean(Command ConstructorVal CoreM Expr Ident InductiveVal Macro Name Syntax TSyntax Term getConstInfoCtor getConstInfoInduct getEnv getStructureFieldsFlattened isInductive isStructure mkIdent quote)
 open Lean.Core(mkFreshUserName)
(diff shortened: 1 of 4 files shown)


KLR/Util/Sexp.lean
```diff
@@ -487,42 +527,47 @@ def toSyntax (tagInfo : TagInfo) : TermElabM Term :=
...
+          | none => pure none
+          | some d => do
+            let c <- Lean.PrettyPrinter.delab d
+            pure (some c)
+          let username := if !localDecl.userName.hasMacroScopes then some localDecl.userName else none
+          match username with
+          | some name => userNames := userNames.push name
+          | none => pure ()
+          let a := mkIdent (← mkFreshUserName `a)
+          binders := binders.push (a, localDecl.type)
```
govereau, line 564: A little bit of a side comment, but I have been thinking about what we should do about `Prop`, that is when `inferType localDecl.type` is `.sort .zero`? For `ToCBOR` I could just  skip the field, but for `FromCBOR`, it is not clear.

govereau (approved):
Great, we need a Meta library.

KLR/Util/Sexp.lean
seanmcl: How would Prop ever show up in FromCBOR?

seanmcl: So we'd need to do arbitrary automated theorem proving during parsing? If so I'd say yes, let's disallow.

PR #218, opened 2025-07-14
chore: Add header script

seanmcl:


interop/klr/peg_parser/action_helpers.c
```diff
@@ -1,3 +1,8 @@
+/*
+Copyright (c) 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Joonwon Choi, Markus De Medeiros, Paul Biberstein, Paul Govereau, Pavel Potapov, Sean McLaughlin
```
govereau, line 4: These files are copied form CPython. Maybe we should skip these?
Actually, you can get all the C files by changing the code generator in `Extract/C.lean`

bin/add_headers
```diff
@@ -0,0 +1,378 @@
+# Copyright (c) 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Released under Apache 2.0 license as described in the file LICENSE.
+# Authors: Joonwon Choi, Markus De Medeiros, Paul Biberstein, Paul Govereau, Pavel Potapov, Sean McLaughlin
```
govereau, line 3: That is a lot of bash code.

govereau (approved):

bin/add_headers
seanmcl: switched to python. q is better at this anyway

PR #226, opened 2025-07-16
Fix useful exception being lost during gather

graebm:
### Issue
I tried to create a kernel, but it failed with the not-very-useful "Unable to fetch NKI function from Python Environment" exception. I need a better exception than that to know what's going wrong

### Diagnosis

This exception is raised if the gather step failed, but no other exception has been raised (code [here](https://github.com/leanprover/KLR/blob/544397526f1f1eb2aa71e34f288427113e844bc9/interop/klr/frontend.c#L40-L43)).

Turns out a more specific exception **was** being raised [here](https://github.com/leanprover/KLR/blob/544397526f1f1eb2aa71e34f288427113e844bc9/interop/klr/gather.c#L701) reporting that an unsupported language feature was being used.

But that exception was being stomped by a later function call to py_fun_name() which always clears any exceptions on the stack (code [here](https://github.com/leanprover/KLR/blob/544397526f1f1eb2aa71e34f288427113e844bc9/interop/klr/gather.c#L151)).

### Solution

Rework some error handling so that we don't call bother calling py_fun_name() if some previous function had failed. I know `goto` gets a bad wrap in modern languages that can clean up after themselves. But C can't clean up after itself, and goto can be a real great tool

### Testing

I made a simple kernel that raised an exception, after this fix I got a useful exception about "unsupported statement" that pointed right at the offending line of code.

### Future work

Revamp gather.c so that:
- any function that truly fails is sure to raise an exception
- we stop work immediately upon a function failing, so that the exception isn't lost. Vs current state where a bunch of functions may be called in a row and then we check if they all succeeded.

Currently, it's pretty wishy washy what any given function is going to with regards to exceptions, which can lead to cases like this.


govereau (approved):
Looks good. Thank you!

PR #227, opened 2025-07-17
[fix]: Add missing KLR instructions

P-bibs:
Follow-up to #214 that adds two operators that were for some reason missing from that PR.
* TensorScalar
* TensorTensor
Also adds dma operators for working with HBM
* DmaHbmLoad
* DmaHbmStore

### DIFF
--- a/KLR/Core/Operators.lean
+++ b/KLR/Core/Operators.lean
@@ -289,7 +289,8 @@ structure AffineSelect where
 
 /- DmaCopy instruction
 
-Generate descriptors to use the DMA to perform a copy.
+Generate descriptors to use the DMA to perform a copy. This is only used
+for sbuf/psum transfers. For interacting with the HBM, use DmaHbmLoad/Store
 
 TODO: this operation is stateful since it uses the DMA queues. How
 do we represent that at the KLR level?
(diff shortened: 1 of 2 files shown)


seanmcl (approved):

KLR/Core/Operators.lean
```diff
@@ -537,7 +537,30 @@ structure TensorReduce where
...
+/- TensorTensor instruction
+-/
+@[serde tag = 172]
+structure TensorTensor where
+  dst : TensorRef
+  src0 : TensorRef
+  src1 : TensorRef
+  op : AluOp
+
+@[serde tag = 173]
```
seanmcl, line 563: these seemingly arbitrary serde tags freak me out a bit. Do we check that we don't have duplicates?

govereau: No good answer here. I started a doc in `docs`, but we need to check this.
I was thinking we could write a meta program that, given a namespace, finds all the types in the namespace with a serde tag and makes sure there are no duplicates.
This unique-in-namespace convention is a proxy for unique-in-a-klr-file. Meaning the important property is that type types that can end up in the same file can't share an id.

P-bibs:
@govereau this has what you need re: HBM if you want to merge into either your branch or main

PR #234, opened 2025-07-22
fix: minor fixes to partial evaluator

govereau:
Mostly changes to naming needed for easy C++ code generation, but also some minor fixes to the new ISA instructions.

### DIFF
--- a/KLR/Core/Operators.lean
+++ b/KLR/Core/Operators.lean
@@ -309,10 +309,10 @@ structure DmaCopy where
 
 Uses the DMA to load a tensor from HBM into SBUF.
 -/
-@[serde tag = 148]
+@[serde tag = 188] -- TODO tag is messed up
 structure DmaHbmLoad where
   dst            : TensorRef
-  src            : HbmTensor
+  src            : TensorHbm
   compute_op     : DgeComputeOp
   dstBoundsCheck : DmaBounds
   srcBoundsCheck : DmaBounds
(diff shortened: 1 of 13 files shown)


KLR/Serde/Basic.lean
```diff
@@ -366,7 +366,7 @@ instance : ToCBOR Int64 where
 instance : FromCBOR UInt64 where
   parse arr :=
     fromFixed arr 0x1b
-      (parseMap UInt16 fun x => x.toUInt64)
+      (parseMap UInt32 fun x => x.toUInt64)
```
ppotapov-aws, line 369: I guess you pulled in my change before I rebased on main

KLR/Core/Operators.lean
```diff
@@ -309,10 +309,10 @@ structure DmaCopy where
 
 Uses the DMA to load a tensor from HBM into SBUF.
 -/
-@[serde tag = 148]
+@[serde tag = 188] -- TODO tag is messed up
```
ppotapov-aws, line 312: Tags had shifted. I guess this is not desirable

ppotapov-aws (approved):

KLR/Core/Operators.lean
govereau: Yup, not great. This is fixed in a follow-up PR.

KLR/Core/Tensor.lean
```diff
@@ -226,6 +249,29 @@ def size (slice : Slice) : Nat :=
 #guard (make! 0 10 (-1)).size == 0
 #guard (make! 10 0 1).size == 0
 
+-- NOTE: The Prop fields count towards the list, but have zero size
+instance : ToCBOR Slice where
```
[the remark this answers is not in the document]
govereau: I have been thinking about this. I think we can do something if there is a `Decidable` instance for the type.

PR #236, opened 2025-07-23
feat: emit operators as side effect instead of returning

ppotapov-aws:


KLR/Trace/ISA.lean
```diff
@@ -218,18 +221,18 @@ nki tensor_tensor_scan
...
 
     if initial.isRight then
       throw "Tensor initializer is not supported"
     let imm := <- initial.getLeft?
 
     let tts : TensorTensorScan := ⟨ dstT, data0T, data1T, op0, op1, rev, imm, .Idle ⟩
-    return .oper [.tensorTensorScan tts]
-
+    Trace.add_stmt $ .oper (.tensorTensorScan tts)
+    return .none
```
seanmcl, line 235: Should this have type `M Unit` if we're always returning none?

govereau: Maybe we should say `Term.none` so it is more clear?

PR #241, opened 2025-07-23
Lowering pass from Access to AccessPattern

markusdemedeiros:
As noted to @govereau: this assumes all tensors are in row-major form. To do this lowering one needs to know how the tensor each Access will be applied to will be laid out in memory. However,
- Accesses are values, so this might not be known until runtime, and 
- it seems that NKI just assumes every tensor is in Row-major form anyways. 
The latter means that we can just assume every access is applied to a tensor in row-major form, and that's what is done in this PR.


ppotapov-aws (approved):

govereau (approved):
Thanks!

PR #246, opened 2025-07-25
chore: operators: check mask and deal with accesses properly

ppotapov-aws:
Outstanding tasks:
 - [x] : Implement ranged select
 - [x] : Go through operations and check LowerAP and tensors functions to match
 - [x] : Add new operations to Extract.C


KLR/Trace/ISA.lean
```diff
@@ -231,386 +224,434 @@ nki tensor_tensor_scan
...
---     if let some m := mask then
---       kwargs := kwargs ++ [⟨ "mask", immediateToValue m ⟩]
---
---     -- TODO If dtype is specified, it should call to cast. Cast is currently not implemented
---
---     let ty := .tensor dst.dtype dst.shape
---     return .expr (.call "tensor_copy_dynamic_dst" args kwargs) ty
+nki tensor_copy_dynamic_src
+ (dst : Access)
+ (src : Access)
```
aws-qieqingy, line 354: I am not sure whether using the default `Access` here is enough. As the name suggest, the src here is dynamically indexed, which is not something the `Access` support.

KLR/Trace/ISA.lean
```diff
@@ -231,386 +224,434 @@ nki tensor_tensor_scan
...
+      dst := .abstract dst
+      src := .abstract src
+      dtype := dtype
+      engine := engine
+    }
+    return .none
 
+nki tensor_copy_dynamic_dst
+ -- kwargs
+ (dst : Access)
```
aws-qieqingy, line 370: Similar to `tensor_copy_dynamic_src`, the `dst` is a dynamic access. The default `Access` class might not be sufficient.

KLR/Trace/ISA.lean
```diff
@@ -231,386 +224,434 @@ nki tensor_tensor_scan
...
+    Trace.add_stmt $ .oper $ .reciprocal {
+      dst := .abstract dst
+      src := .abstract data
+      dtype := dtype
+    }
     return .none
 
 nki iota
  (dst: Access)
  (_expr : Int) -- TODO: Placeholder. Figure out this type
```
aws-qieqingy, line 418: This needs be an expression out of the `nl.mgrid`, `nl.arange`, or something that is as expressive as the APPairs.

KLR/Trace/ISA.lean
```diff
@@ -231,386 +224,434 @@ nki tensor_tensor_scan
...
+          | .inr t => .tile $ .abstract t
+        thresholdType := .KeepRate
+        dtype         := dtype,
+    }
+    return .none
 
-    Trace.add_stmt $ .oper (.dropout ⟨ dstT, dataT, .KeepRate , prob  ⟩)
+nki affine_select
+ (dst: Access)
+ (_pred : Int) -- TODO Placeholder. Figure out this type
```
aws-qieqingy, line 451: Similar to Iota, this needs to be something as expressive as an ApPir.

KLR/Trace/ISA.lean
```diff
@@ -231,386 +224,434 @@ nki tensor_tensor_scan
...
+      compOp0 := comp_op0,
+      compOp1 := comp_op1,
+      bound0 := .abstract bound0,
+      bound1 := .abstract bound1,
+      rangeStart := range_start,
+      onTrueTile := .abstract on_true_tile,
+      onFalseValue := on_false_value,
+      dtype := dtype
+    }
+    return .none
```
aws-qieqingy, line 498: We are missing `select_reduce` and `sequence_bounds` here. These were merged in recently and they are not released yet. Please take a look at the internal API page.

KLR/Trace/ISA.lean
```diff
@@ -231,386 +224,434 @@ nki tensor_tensor_scan
...
 
 nki memset
  (dst: Access)
  (shape : Shape)
```
aws-qieqingy, line 501: Memset does not need a shape anymore.

KLR/Trace/ISA.lean
```diff
@@ -231,386 +224,434 @@ nki tensor_tensor_scan
...
+      dst := .abstract dst,
+      src := .abstract src,
+      compute_op := op,
+      oobMode := match mode with
+        | 0 => .enable
+        | 1 => .disable
+        | _ => .disable,
+      dgeMode := dge_mode,
+  }
   return .none
```
aws-qieqingy, line 588: dma_transpose is still missing

KLR/Trace/ISA.lean
```diff
@@ -231,386 +224,434 @@ nki tensor_tensor_scan
...
+ (dtype: Option Dtype := none)
+ (mask: Option Immediate := none) := do
+    if mask.isSome then throw maskNotSupported
+    Trace.add_stmt $ .oper $ .shuffle {
+      dst := .abstract dst,
+      src := .abstract src,
+      shuffleMask := shuffle_mask,
+      dtype := dtype,
+    }
     return .none
```
aws-qieqingy, line 657: How do we want to handle the stuff in the private API namespace, a few of them are used in the production today.

KLR/Trace/ISA.lean
```diff
@@ -61,164 +63,157 @@ nki nc_matmul
...
+ -- kwargs
+ (reduce_op : Option AluOp := none)
+ (reduce_res : Option Access := none)
+ (bias : List Immediate := [])
+ (scale : Sum Immediate Access := .inl (.float 1.0))
+ (mask : Option Immediate := none)
+ (dtype : Option Dtype := none) := do
+    if mask.isSome then
+      throw maskNotSupported
+    Trace.add_stmt $ .oper $ .activationReduce {
```
aws-qieqingy, line 141: ActivationReduce is strictly a subset of the Activate instruction. We should not have a dedicated ActivationReduce instruction.

interop/klr/klir_ast.hpp
```diff
@@ -489,38 +555,43 @@ struct MatMul final {
...
+  Ptr<TensorRef> onTrueTile;
+  Ptr<Immediate> onFalseValue;
+  Option<Dtype> dtype;
 };
 
 struct ScalarTensorTensor final {
   Ptr<TensorRef> dst;
-  Ptr<TensorRef> src0;
-  Ptr<TensorRef> src1;
+  Ptr<TensorRef> data;
```
aws-qieqingy, line 580: In STT, the data can be a const, so I think data needs to be an Operand. but the src0 and src1 are tensors, so they need to be TensorRef. Please take a look at: https://awsdocs-neuron.readthedocs-hosted.com/en/latest/general/nki/api/generated/nki.isa.scalar_tensor_tensor.html

KLR/Trace/ISA.lean
ppotapov-aws: Collapsed

KLR/Trace/ISA.lean
ppotapov-aws: Removed shape

KLR/Trace/ISA.lean
```diff
@@ -54,561 +57,641 @@ def accumCmdToValue (ac : AccumCmd) : Value :=
...
+      src := .abstract data,
+      dtype := dtype
+    }) name
     return .none
 
 nki iota
  (dst: Access)
- (_expr : Int) -- TODO: Placeholder. Figure out this type
+ (name : String)
+ (expr : Int)
```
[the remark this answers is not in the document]
govereau: I think it needs to be `DataPattern` going by the HW.

KLR/Eval.lean
```diff
@@ -24,7 +24,7 @@ open Lean(AssocList)
 open StateT(lift)
```
govereau, line 24: Are the changes to Eval still needed?

ppotapov-aws: Somehow I missed that one. I changed Eval reflects order of things a bit better

govereau (approved):
Looks good.
Let's merge this before it manages to get bigger.

PR #252, opened 2025-08-01
feat: build klr binary locally with docker

ppotapov-aws:
It is useful for building backend compiler to be able to build locally off specified branch and replace binary in place without trying to fetch and unpack artifacts


ppotapov-aws:
Closing due to different way of distributing binaries that don't depend on lean

PR #255, opened 2025-08-04
fix: chained comparison should not short-circuit

govereau:
In Python, a chained comparison does not short-circuit. For example, if you write the expression:

  1 < 0 < f()

then the function f will be called. This is different from logical and/or which does short-circuit. We also have the possibility of logical and/or on tensors, which does not short-circuit. This change introduces two new expression forms `conj` and `disj` which are used to represent Python short-circuit and/or. The normal land and lor binary operator do not short-circuit. This gives us what we need to lower logical operators, tensor operators, and chained comparison.


seanmcl (approved):
Nuts. Do we see chained comparison in kernels? 

Hard to see exactly what is going on without examples, e.g. of the effect of the let binding, but lgtm.

govereau:
> Nuts. Do we see chained comparison in kernels?

Nope, we never see it. I guess that is a good thing since it was wrong before.

PR #273, opened 2025-08-11
fix: Switch Enum from Elab to Macro, fix private toUInt8

seanmcl:
Macros are lighter weight, and I was able to debug this faster. If an inductive is private, You need private on any method you want to use postfix.

### DIFF
--- a/KLR/Util/Enum.lean
+++ b/KLR/Util/Enum.lean
@@ -21,11 +21,9 @@ import Util.Sexp
 
 open KLR.Util.Hex(encode)
 open KLR.Util(FromSexp ToSexp)
-open Lean(FromJson Json Syntax TSyntax TSyntaxArray ToJson mkIdent fromJson? toJson)
-open Lean.Elab.Command(CommandElab CommandElabM elabCommand liftTermElabM)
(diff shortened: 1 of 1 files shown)


KLR/Util/Enum.lean
```diff
@@ -57,15 +55,19 @@ syntax item := Lean.Parser.Command.ctor (":=" num)?
...
 | `($modifiers:declModifiers enum $name:ident where $items:item* $dcs:optDeriving) => doit modifiers name items dcs
-| _ => throwError "invalid enum syntax"
+| _ => Macro.throwUnsupported
 where
   doit (modifiers : TSyntax `Lean.Parser.Command.declModifiers)
        (name : TSyntax `ident)
        (items : TSyntaxArray `KLR.Util.item)
-       (dcs : TSyntax `Lean.Parser.Command.optDeriving) : CommandElabM Unit := do
+       (dcs : TSyntax `Lean.Parser.Command.optDeriving) : MacroM (TSyntax `command) := do
+    let priv := ((modifiers.raw.getArg 2).getArg 0).getArg 0
```
govereau, line 67: Are the modifiers always in a specific order? Why is it 2, 0, 0 ?

PR #274, opened 2025-08-12
chore: Add Vector to NumBytes, ToSexp, FromSexp

seanmcl:
These are the easy cases. FromBytes is coming.

### DIFF
--- a/KLR/Util/Sexp.lean
+++ b/KLR/Util/Sexp.lean
@@ -92,6 +92,9 @@ instance [ToSexp a] : ToSexp (List a) where
 instance [ToSexp a] : ToSexp (Array a) where
   toSexp e := toSexp e.toList
 
+instance [ToSexp a] : ToSexp (Vector a n) where
+  toSexp e := toSexp e.toList
+
 instance [ToSexp a] [ToSexp b] : ToSexp (a × b) where
   toSexp p := list [toSexp p.1, toSexp p.2]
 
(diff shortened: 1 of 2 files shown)


KLR/Util/Sexp.lean
```diff
@@ -191,6 +194,20 @@ instance [FromSexp a] : FromSexp (List a) where
...
   | list es => es.mapM (@fromSexp? a _)
 
+instance [FromSexp a] : FromSexp (Vector a n) where
+  fromSexp?
+  | atom a => throw s!"can't parse list from atom: {a}"
+  | list es => do
+    let len := es.length
+    if H : len != n then throw s!"Wrong size list ({len}) for Vector of size {n}" else
+    let es' := es.toArray
+    let arr <- es'.mapM' (@fromSexp? a _)
```
govereau, line 204: Since you are in an error monad, maybe this is better?
```
if H: arr.size = n then
  return .mk arr H
else
  throw "wrong size"
```

govereau (approved):

KLR/Util/Sexp.lean
seanmcl: Hmm, I agree it is simpler. Seems sad to have a dead else branch when it's provably impossible. I guess we need to find a balance when most of what we do is not verified but once in a while we need verified code. Replaced the code, but it's another item to think about for the style guide.

PR #275, opened 2025-08-12
feat: add checking for out-of-bounds slice patterns

govereau:


KLR/Trace/NKI.lean
```diff
@@ -104,6 +104,25 @@ private def lookupName (name : Name) : Trace Term := do
...
+Currently this implementation will silently clip out-of-bounds slice patterns,
+similar to Python and numpy. However, to aide developers we will (try to) warn
+or fail if we detect an overflow. This function is applied to function
+arguments.
+-/
+private def checkAccess (t : Term) (warnOnly : Bool := true) : Trace Unit := do
+  match t with
+  | .expr (.value (.access (.basic b))) _ => do
+    let shape <- b.shape
+    let shape := shape.parDim :: shape.freeDims
```
seanmcl, line 119: Maybe there should be a utility for this, `shape.toList` or something

KLR/Trace/NKI.lean
```diff
@@ -104,6 +104,25 @@ private def lookupName (name : Name) : Trace Term := do
...
 
+/-
+Best effort checks for slice overflow
+
+Currently this implementation will silently clip out-of-bounds slice patterns,
+similar to Python and numpy. However, to aide developers we will (try to) warn
+or fail if we detect an overflow. This function is applied to function
+arguments.
+-/
+private def checkAccess (t : Term) (warnOnly : Bool := true) : Trace Unit := do
```
seanmcl, line 115: All this uses is `b`. I wonder if we should be checking it earlier, like when you make the `b` in the first place.

KLR/Trace/NKI.lean
```diff
@@ -104,6 +104,25 @@ private def lookupName (name : Name) : Trace Term := do
...
+  match t with
+  | .expr (.value (.access (.basic b))) _ => do
+    let shape <- b.shape
+    let shape := shape.parDim :: shape.freeDims
+    for (dim, ndx) in shape.zip b.indexes do
+      if dim > ndx.size then
```
seanmcl, line 121: Should this be `dim < ndx.size`? My reasoning: `ndx.size` is the number of elements you want. It's more than 1 if you have a slice, and you shouldn't be able to grab more elements than size of the dimension.

govereau: Oh, right! oops.

KLR/Trace/NKI.lean
govereau: Funny, there is already a `Shape.toList`, I just didn't remember. Will switch to that.

KLR/Trace/NKI.lean
govereau: Right, thanks. I clicked a button on the github website this morning and it did something weird to my branches, so I tried clicked a "revert" button and got a huge mess. I think it is all fixed now.

seanmcl (approved):

PR #276, opened 2025-08-12
fix: add BirAccessPattern to extraction

govereau:


KLR/Extract/Extract/C.lean
```diff
@@ -198,6 +198,7 @@ def klrAST: MetaM (List LeanType) := do
     `KLR.Core.AccessBasic,
     `KLR.Core.APPair,
     `KLR.Core.AccessPattern,
+    `KLR.Core.BirAccessPattern,
```
seanmcl, line 201: sort

seanmcl (approved):

PR #280, opened 2025-08-13
feat: use variable name if no name given in ndarray

govereau:


KLR/Trace/Term.lean
```diff
@@ -517,7 +517,9 @@ nki ndarray
     | some (par, free) => (some par, some free)
     | none => (none, none)
     let address := { memory, parSize, freeSize, parOffset, freeOffset : Address }
-    let name := name.getD "tensor"
+    let name <- match name with
+      | none => pure (<- genName `tensor).toString
+      | some x => pure x
```
aws-qieqingy, line 522: The logic here is insufficient if `ndarray` is called in a loop. We need to append the ID to deduplicate.

aws-qieqingy (changes requested):

aws-qieqingy (approved):

PR #281, opened 2025-08-13
Add compilation passes HLO->TGR->K3->K2->K1

P-bibs:
Adds code to compile HLO programs to a variant of KLR. See internal documents for more details.

The broad breakdown is as follows:

* HLO: the source language emitted by JAX, we don't have control over this. A graph of tensor operations.
* TGR: a language designed to abstract over the different tensor graphs produced by deep learning frameworks like PyTorch and JAX. The operator set should be small, and each operator should have very simple, elemental semantics. This graph will be of greater or equal size to the HLO graph, since each HLO operation decomposes into one or more TGR operators.
* K3: K3 is the first language that uses the operators that resemble nki.isa (KLR). The process of going from TGR to K3 requires instruction selection, to map the numpy-like TGR operators to isa-like KLR operators. Other than that, K3 is still a tensor graph of arbitrarily large tensors
* K2: K2 tiles each operator in K3 so that explicit hbm->sbuf movement is included and operators only are applied to tiles that are loaded into SBUF. However, tiles do not have explicit locations allocated.
* K1: K1 looks largely similar to K2, except an allocator has provided a concrete location for each tile and all scalar expressions have been unfolded into register operations, with registers also being allocated.

To run an example compilation:
```
lake exe klr compile-hlo add-opt.txt -o compiled/add.py
lake exe klr compile-hlo dotproductattention-opt.txt -o compiled/dpa.py
```
[add-opt.txt](https://github.com/user-attachments/files/21921234/add-opt.txt)
[dotproductattention-opt.txt](https://github.com/user-attachments/files/21921235/dotproductattention-opt.txt)

**Note:** The changes in the `TGR` dir have already been checked in a separate code review. The remainder of the changes are largely in the `K` directory and amount to 1500 sloc, after excluding KLR/K/Operators.lean, which is a near-duplicate of KLR/Core/Operators.lean, which has already been reviewed.


P-bibs:
Sure, do you mean for review purposes? I was under the impression I would squash these all into one commit once it's time to merge.

govereau:
> Sure, do you mean for review purposes? I was under the impression I would squash these all into one commit once it's time to merge.

Oh, I though you were proposing to merge 40 commits. Nevermind.

govereau (approved):

KLR/Core/Tensor.lean
```diff
@@ -598,9 +630,14 @@ parQuadrant─►96│    ┌───────┐│        │
 structure TensorSram where
   name : String
   dtype : Dtype
+  memory : SramMemory
```
P-bibs, line 633: I added this `memory` member for the tensor graph compilation path since it uses `TensorSram` directly instead of through `TensorRef`, but it occurs to me that now if you use `TensorRef` the discriminant is redundant with this memory field. Should I move this TensorSram definition into the `K` folder and make it its own thing?

govereau: If you want to change it, go ahead... You can merge if you want as well.

PR #292, opened 2025-08-14
fix: add mgrid mul with integer

aws-qieqingy:


KLR/Trace/Term.lean
```diff
@@ -159,8 +159,9 @@ def termOp : BinOp -> Term -> Term -> Trace Term
   | .mul, .tuple  l,          .expr (.value  v) _
   | .mul, .expr (.value v) _, .tuple l  => return .tuple (<- mulseq l v)
   -- mgrid
-  | .add, .expr (.value (.int i)) _, .mgItem a b
-  | .add, .mgItem a b, .expr (.value (.int i)) _ => return .mgItem (a+i) (b+i)
+  | .add, .expr (.value (.int i)) _, .mgItem a b step
+  | .add, .mgItem a b step, .expr (.value (.int i)) _ => return .mgItem (a+i) (b+i) step
+  | .mul, .mgItem a b step, .expr (.value (.int i)) _ => return .mgItem a b (i*step)
```
govereau, line 164: Do you need `b * i`? Meaning `b` is the end so it needs to be moved.
Either that or the implementation of `mgrid` needs to be changed.

KLR/Trace/Term.lean
```diff
@@ -433,9 +434,9 @@ def mgrid (indexes : List Term) : Err Term := do
       if b == none then
         throw "size not specified"
       let b := b.get!
-      if c != none && c != some 1 then
-        throw "step size must be 1"
-      l := l ++ [.mgItem a b]
+      match c with
+      | none => l := l ++ [.mgItem a b 1]
+      | some step => l := l ++ [.mgItem a b step]
```
govereau, line 439: Now you can say `c.getD 1` instead of the match. "get with default in case of none"

govereau (changes requested):

PR #293, opened 2025-08-15
chore: Update to 4.22

seanmcl:
There is some problem with conflicting versions that breaks the cli "...foo" syntax. I commented it out and imagine it'll get fixed once all our repos are on the same version.


govereau (approved):

seanmcl:
OK, this is fixed. You really need lake update in all subdirs with a lakefile, which is annoying. I'll add a script that does this, and also start moving lakefile instructions up to the top level. Will try to have a single lakefile per repo.

seanmcl:
Ug! only fixed locally. Lake is not quite working as expected/desired here.

PR #310, opened 2025-08-22
Chore: align better with ISA

ppotapov-aws:


KLR/Trace/ISA.lean
```diff
@@ -745,8 +748,8 @@ nki builtin.isa.nc_match_replace8
 
 
 nki builtin.isa.nc_stream_shuffle
- (src : Access)
  (dst : Access)
+ (src : Access)
```
waahm7, line 752: This was already aligned with the NKI API https://awsdocs-neuron.readthedocs-hosted.com/en/latest/general/nki/api/generated/nki.isa.nc_stream_shuffle.html. I was proposing that we just change the API to make it consistent with all the other APIs.

waahm7 (approved):

KLR/Trace/ISA.lean
ppotapov-aws: Happy to go either way. I think uniformity is best from user perspective

govereau (approved):

PR #328, opened 2025-08-27
Feat: better error messaging during tracing

ppotapov-aws:
When we encounter unexpected expression we should let user know which expression was not supported.

### DIFF
--- a/KLR/Trace/Term.lean
+++ b/KLR/Trace/Term.lean
@@ -45,7 +45,7 @@ private def mulseq (l : List a) : Term -> Err (List a)
   | .bool false => return []
   | .bool true  => return l
   | .int i      => return List.flatten $ List.replicate i.toNat l
-  | _           => throw "invalid multiply"
+  | t           => throw s!"can't multiply sequence by '{Term.kindStr t}'"
 
 -- Binary operators on constants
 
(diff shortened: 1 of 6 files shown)


govereau (approved):

KLR/Trace/Term.lean
```diff
@@ -163,7 +163,7 @@ private def valueLt : Term -> Term -> Trace Bool
   | .bool b, .int i => return (if b then 1 else 0) < i
   | .int i, .bool b => return i < (if b then 1 else 0)
   -- errors
-  | _, _ => throw "unsupported comparison"
+  | l, r => throw s!"unsupported comparison between {Term.kindStr l} and {Term.kindStr r}"
```
graebm, line 166: Probably worth throwing quotes around all these strings, to make it clear what is part of the error message, and what it is part of the string-from-the-outside-world. Especially valuable if that string-from-the-outside-world ends up being blank
```suggestion
  | l, r => throw s!"unsupported comparison between '{Term.kindStr l}' and '{Term.kindStr r}'"
```

I checked out what Python does, and it quotes things:
```
>>> 3 + "a"
Traceback (most recent call last):
  File "<python-input-0>", line 1, in <module>
    3 + "a"
    ~~^~~~~
TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

ppotapov-aws: Accepted. Neat suggestion feature

graebm: I meant, for ALL these errors messages, we should quote the strings

PR #329, opened 2025-08-27
Fix SharedConstant Metadata

waahm7:
- Fix the path for the shared constant files
- Include shared constant files path in the metadata.

### DIFF
--- a/KLR/Compile.lean
+++ b/KLR/Compile.lean
@@ -67,13 +67,18 @@ structure KernelInfo where
   name : String
   inputs : List TensorInfo
   outputs : List TensorInfo
+  sharedConstants : List Core.SharedConstantFile
   deriving ToJson
+
+private def outfolder (outfile : String) : String :=
+  let path := FilePath.mk outfile
+  (path.parent.map (·.toString)).getD "."
 
 -- reads srcPythonAstFileName, writes dstKlrFileName, returns kernel info as string of json
 @[export klr_frontend_trace]
 def frontend_trace (srcPythonAstFileName dstKlrFileName : String) : IO String := do
   let kernel <- KLR.File.readKLRFile srcPythonAstFileName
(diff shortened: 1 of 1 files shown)


ppotapov-aws (approved):

KLR/Compile.lean
```diff
@@ -67,13 +67,20 @@ structure KernelInfo where
   name : String
   inputs : List TensorInfo
   outputs : List TensorInfo
+  sharedConstants : List Core.SharedConstantFile
   deriving ToJson
 
+private def outfolder (outfile : String) : (Option String) :=
```
govereau, line 73: Return type can be just `String`, since you never return `none`.

KLR/Compile.lean
```diff
@@ -67,13 +67,20 @@ structure KernelInfo where
...
+  let path := FilePath.mk outfile
+  match path.parent with
```
govereau, line 75: same as: `path.getD "."`

KLR/Compile.lean
```diff
@@ -67,13 +67,20 @@ structure KernelInfo where
...
+  | none => some "."
+  | some parent => some parent.toString
+
 -- reads srcPythonAstFileName, writes dstKlrFileName, returns kernel info as string of json
 @[export klr_frontend_trace]
 def frontend_trace (srcPythonAstFileName dstKlrFileName : String) : IO String := do
   let kernel <- KLR.File.readKLRFile srcPythonAstFileName
-  let kernel <- compilePython kernel none
+  let kernel <- compilePython kernel $ outfolder dstKlrFileName
```
govereau, line 83: Since there is not way to fail, we can change the argument type of `compilePython`.

govereau (approved):

KLR/Compile.lean
waahm7: Thanks, updated.

PR #332, opened 2025-08-28
fix: improve error messages for unsupported stmt kinds

graebm:
Error messages must be descriptive and user-actionable.

I couldn't think of user-actionable advice for some of these (with, del, type, global, nonlocal) because I wasn't sure if we'd be supporting these soon or not. Please comment on the `TODO: ADVICE` lines with your suggestion.

### DIFF
--- a/interop/klr/gather.c
+++ b/interop/klr/gather.c
@@ -1040,8 +1040,71 @@ static struct Python_Stmt* stmt(struct state *st, struct _stmt *python) {
 
     // TODO: do we need with?
     case With_kind:
+      syntax_error(st, "NKI does not support 'with' statements at this time.");
+      res = NULL;
+      break;
+
+    case FunctionDef_kind:
+      syntax_error(st, "NKI does not support inner function definitions. Move function definition outside this function.");
+      res = NULL;
+      break;
+
+    case ClassDef_kind:
+      syntax_error(st, "NKI does not support 'class' definitions within a function. Move class definition outside this function.");
+      res = NULL;
+      break;
+
+    case Delete_kind:
+      syntax_error(st, "NKI does not support 'del' statements at this time.");
+      res = NULL;
+      break;
+
+    case TypeAlias_kind:
+      syntax_error(st, "NKI does not support 'type' statements at this time.");
+      res = NULL;
+      break;
+
+    case AsyncFunctionDef_kind:
+    case AsyncFor_kind:
+    case AsyncWith_kind:
+      syntax_error(st, "NKI does not support 'async'. Use only synchronous functions within kernels.");
+      res = NULL;
+      break;
+
+    case Match_kind:
+      syntax_error(st, "NKI does not support 'match' statements at this time. Use 'if/elif' or dict lookups instead.");
+      res = NULL;
+      break;
+
+    case Raise_kind:
+      syntax_error(st, "NKI does not support 'raise' statements. Use 'if/else' control flow within kernels, or 'assert' for fatal errors.");
+      res = NULL;
+      break;
+
+    case Try_kind:
+    case TryStar_kind:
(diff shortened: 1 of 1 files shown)


interop/klr/gather.c
```diff
@@ -1040,8 +1040,71 @@ static struct Python_Stmt* stmt(struct state *st, struct _stmt *python) {
...
+      res = NULL;
+      break;
+
+    case ClassDef_kind:
+      syntax_error(st, "NKI does not support 'class' statements. Move class definition outside the kernel.");
+      res = NULL;
+      break;
+
+    case Delete_kind:
+      syntax_error(st, "NKI does not support 'del' statements at this time. TODO: ADVICE");
```
govereau, line 1058: No advice here. We don't have environments like Python, so there is nothing like `del`

interop/klr/gather.c
```diff
@@ -1040,8 +1040,71 @@ static struct Python_Stmt* stmt(struct state *st, struct _stmt *python) {
...
+      res = NULL;
+      break;
+
+    case TypeAlias_kind:
+      syntax_error(st, "NKI does not support 'type' statements at this time. TODO: ADVICE");
```
govereau, line 1063: We could just ignore these, I mean Python just ignores them?

interop/klr/gather.c
```diff
@@ -1040,8 +1040,71 @@ static struct Python_Stmt* stmt(struct state *st, struct _stmt *python) {
...
+      break;
+
+    case Import_kind:
+    case ImportFrom_kind:
+      syntax_error(st, "NKI does not support 'import' statements. Move 'import' outside the kernel function.");
+      res = NULL;
+      break;
+
+    case Global_kind:
+      syntax_error(st, "NKI does not support 'global' statements at this time. TODO: ADVICE");
```
govereau, line 1097: Same as above, no environments so no global.

interop/klr/gather.c
```diff
@@ -1040,8 +1040,71 @@ static struct Python_Stmt* stmt(struct state *st, struct _stmt *python) {
...
+      res = NULL;
+      break;
+
+    case Nonlocal_kind:
+      syntax_error(st, "NKI does not support 'nonlocal' statements at this time. TODO: ADVICE");
```
govereau, line 1102: Same, environments.

govereau (approved):

interop/klr/gather.c
graebm: I'm not really sure how to ignore them, so leaving it as is for now

interop/klr/gather.c
```diff
@@ -1040,8 +1040,71 @@ static struct Python_Stmt* stmt(struct state *st, struct _stmt *python) {
...
+      res = NULL;
+      break;
+
+    case FunctionDef_kind:
+      syntax_error(st, "NKI does not support 'def' statements. Move function definition outside the kernel.");
+      res = NULL;
+      break;
+
+    case ClassDef_kind:
+      syntax_error(st, "NKI does not support 'class' statements. Move class definition outside the kernel.");
```
aws-qieqingy, line 1053: Same here, "NKI does not support 'class' definition within functions....."

interop/klr/gather.c
```diff
@@ -1040,8 +1040,71 @@ static struct Python_Stmt* stmt(struct state *st, struct _stmt *python) {
...
+
+    case Try_kind:
+    case TryStar_kind:
+      syntax_error(st, "NKI does not support 'try' statements. Use 'if/else' control flow within kernels.");
+      res = NULL;
+      break;
+
+    case Import_kind:
+    case ImportFrom_kind:
+      syntax_error(st, "NKI does not support 'import' statements. Move 'import' outside the kernel function.");
```
aws-qieqingy, line 1092: Same here, "NKI does not support 'import' within functions, ...."

aws-qieqingy (approved):

PR #339, opened 2025-08-29
fix: consistent naming of "psum"

graebm:
Previously, it was called "pmem" in some places. This makes it consistent.

### DIFF
--- a/KLR/KLR/Basic.lean
+++ b/KLR/KLR/Basic.lean
@@ -32,7 +32,7 @@ inductive Engine where
 
 -- Memory types
 inductive Memory where
-  | hbm | sbuf | pmem
+  | hbm | sbuf | psum
   deriving Repr, BEq
 
 /-
--- a/interop/neuronxcc/nki/language/__init__.py
+++ b/interop/neuronxcc/nki/language/__init__.py
@@ -10,7 +10,7 @@ def ndarray(shape, dtype, buffer=None, name=''):
   elif buffer == nki.language.sbuf:
     buffer = sbuf
   elif buffer == nki.language.psum:
-    buffer = pmem
+    buffer = psum
   else:
     assert False, "invalid buffer argument"
   return buffer.view(dtype, shape, name)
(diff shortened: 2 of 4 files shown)


KLR/KLR/Basic.lean
```diff
@@ -32,7 +32,7 @@ inductive Engine where
 
 -- Memory types
 inductive Memory where
-  | hbm | sbuf | pmem
+  | hbm | sbuf | psum
   deriving Repr, BEq
```
govereau, line 36: The definition used by the extracted C code is here:
https://github.com/leanprover/KLR/blob/a88423067cdfff6e905fc07f425ee28a69001354/KLR/Core/Tensor.lean#L29-L33
Should update both.

interop/neuronxcc/nki/language/__init__.py
```diff
@@ -10,7 +10,7 @@ def ndarray(shape, dtype, buffer=None, name=''):
   elif buffer == nki.language.sbuf:
     buffer = sbuf
   elif buffer == nki.language.psum:
-    buffer = pmem
+    buffer = psum
```
govereau, line 13: This is controlled by:
https://github.com/leanprover/KLR/blob/359e9154b25b01d35116b45632f67ec4e8e723ad/KLR/Trace.lean#L35

KLR/KLR/Basic.lean
govereau: We should also regenerate the C++ definitions. In this case it doesn't matter, but for consistency we probably want to C++ code to use the name "psum" also.

govereau (dismissed):

interop/neuronxcc/nki/language/__init__.py
graebm: not sure if you're saying I should change something, or just sharing info

KLR/KLR/Basic.lean
govereau: Oh, sorry. I thought you were changing `psum -> pmem`, nvm.

govereau (approved):

PR #342, opened 2025-09-04
Feat: implement fromNKI? using c++

ppotapov-aws:
Implement `FromNKI?` functionality defined in lean using c++
Notes:
- We don't have a tensor library. I will defer until later with tensor implementation
- The access definition doesn't match lean definition. I can follow up here or on separate PR

### DIFF
--- a/interop/klr/fromNKI.hpp
+++ b/interop/klr/fromNKI.hpp
@@ -35,7 +35,7 @@ Error TraceError(std::string msg) {
 
 struct Term;
 
-struct Builtin final { std::string name; std::optional<Term*> term; };
+struct Builtin final { std::string name; std::optional<std::shared_ptr<Term>> term; };
 struct Ref final {
      std::string name;
      enum Kind {


interop/klr/fromNKI.hpp
```diff
@@ -0,0 +1,511 @@
...
+#include <variant>
+#include <string>
+#include <vector>
+#include <optional>
+
+namespace Trace {
+
+struct Error {
+    std::string msg;
+    Error(std::string&& m) : msg(std::move(m)) {}
```
graebm, line 14: If you just take the string by value, it's way easier for users (and just as efficient). They can call it like: `Error("oh no")` instead of `Error(std::move(std::string("oh no")))`. I see you discovered that and made a `TraceHelper(std::string msg)` helper function later to get the same result as what I'm suggesting here. You wouldn't need that helper anymore

The only advantage of the "r-value only" version you made here is that users are forced to call move() themselves and can't accidentally do a copy instead. But you can see the downside, as you had to make a helper function to work around the clunkiness.

```suggestion
    Error(std::string m) : msg(std::move(m)) {}
```

interop/klr/fromNKI.hpp
```diff
@@ -0,0 +1,511 @@
...
+
+template<typename T>
+T get_ok(const TraceResult<T>& result) {
+    return std::get<T>(result);
+}
+
+Error TraceError(std::string msg) {
+    return Error{std::move(msg)};
+}
+
```
graebm, line 34: Don't need this if you make change to Error constructor I suggested above
```suggestion
```

interop/klr/fromNKI.hpp
```diff
@@ -0,0 +1,511 @@
...
+namespace Trace {
+
+struct Error {
+    std::string msg;
+    Error(std::string&& m) : msg(std::move(m)) {}
+};
+
+// Simple error handling
+template<typename T>
+using TraceResult = std::variant<T, Error>;
```
graebm, line 19: Don't need it in this PR, but I'd love a proper Result class.
Or use c++ exceptions.

interop/klr/fromNKI.hpp
```diff
@@ -0,0 +1,511 @@
...
+
+template<typename T>
+bool is_ok(const TraceResult<T>& result) {
+    return std::holds_alternative<T>(result);
+}
+
+template<typename T>
+T get_ok(const TraceResult<T>& result) {
+    return std::get<T>(result);
```
graebm, line 28: This is returning a copy of T. Because it takes result by reference, but return a T by value.

We might want it to be more like [std::get for variant](https://en.cppreference.com/w/cpp/utility/variant/get.html) and [std::expected::value()](https://en.cppreference.com/w/cpp/utility/expected/value.html). Returning either a reference, or rvalue, to what's inside

```
template<typename T>
T& get_ok(const TraceResult<T>& result) {
    return std::get<T>(result);
}

template<typename T>
T&& get_ok(TraceResult<T>&& result) {
    return std::get<T>(std::move(result));
}
```

graebm (commented):
I stopped an 1/8 of the way in to this review. But here's what I have

PR #351, opened 2025-09-15
feat: add basic infrastructure for classes

govereau:
This change sets up the basic infrastructure for classes and dictionaries. First, the python extension library has been converted to directly construct Lean objects, and we are bypassing the serialization between Python and Lean. The gather step has also been modified to collect top-level class definitions and class values passed as arguments.

In addition, the Python, NKI, and Term types have been updated with variants for classes, objects, methods, and dictionaries. Full tracing support for classes and dictionaries is not yet implemented.


KLR/NKI/Basic.lean
```diff
@@ -166,15 +168,27 @@ instance : BEq Fun where
...
 @[serde tag = 14]
+structure Class where
+  name : String
+  fields : List Keyword
+  methods : List Fun
+  deriving FromCBOR, FromJson, FromSexp, Repr, ToCBOR, ToJson, ToSexp
+
+-- Names are fully qualified and unique
+instance : BEq Class where
+  beq l r := l.name == r.name
```
ppotapov-aws, line 179: Is the name fully qualified here? Will we have conflicts with classes with the same name equal to each other? I.e
```python
# foo/a.py
class A:
  pass
# bar/b.py
class A:
  pass

foo.A() == bar.A()
```

KLR/Trace/Python.lean
```diff
@@ -47,6 +47,9 @@ private partial def termStr : Term -> Trace String
   | .builtin name _ => return name.toString
   | .ref name _ => do termStr (<- lookup name)
   | .source f => return f.name
+  | .cls c => return s!"<class {c}>"
```
ppotapov-aws, line 50: Should we be consistent with naming here. I.e `<class {c}>`, `<object {c}>` or vice versa?

KLR/Trace/Python.lean
```diff
@@ -47,6 +47,9 @@ private partial def termStr : Term -> Trace String
...
+  | .object c _ => return s!"<{c} object>"
+  | .method .. => return "method"
```
ppotapov-aws, line 52: Should we print the name here? Python gives name of the method and class:
```
In [1]:
   ...: class A:
   ...:     def __init__(self, a):
   ...:         self.a = a
   ...:     def foo():
   ...:         print(a)
   ...:

In [2]: A
Out[2]: __main__.A

In [3]: A.foo
Out[3]: <function __main__.A.foo()>

In [4]: a = A(1)

In [5]: a.foo
Out[5]: <bound method A.foo of <__main__.A object at 0x10407ded0>>
```

ppotapov-aws (approved):

KLR/Trace/Python.lean
govereau: Yeah, we should go through all of the string conversions at some point. It is a minor detail so I will ignore this for now.

KLR/Trace/Python.lean
govereau: The name here is the name of the class, but I think I changed this in a later PR. Will revisit.

PR #362, opened 2025-09-18
feat: add support for f-strings

govereau:
This change adds support for Python-style f-strings. This implementation does not support format specifiers, but it does support the repr conversion tag.  In the case of repr-conversion, we use the Lean repr format.

### DIFF
--- a/KLR/NKI/Simplify.lean
+++ b/KLR/NKI/Simplify.lean
@@ -213,6 +213,13 @@ private def expr' (e' : Python.Expr') : Simplify Expr' :=
       return .call f args kws
   | .starred .. => throw "tuple expansion is not supported"
   | .object c fs => return rewriteObj c.toName (<- keywords fs)
+  | .format e r => do
+      let r <- match r with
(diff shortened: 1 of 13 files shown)


KLR/NKI/Simplify.lean
```diff
@@ -213,6 +213,13 @@ private def expr' (e' : Python.Expr') : Simplify Expr' :=
       return .call f args kws
   | .starred .. => throw "tuple expansion is not supported"
   | .object c fs => return rewriteObj c.toName (<- keywords fs)
+  | .format e r => do
+      let r <- match r with
+        | none | some 97 | some 115 => pure false  -- none or 'a' or 's'
+        | some 114 => pure true -- 'r'
+        | some c => throw s!"unsupported conversion sspecifier {c}"
```
ppotapov-aws, line 220: got an extra s in sspecifier 🐍

ppotapov-aws (approved):

KLR/NKI/Simplify.lean
govereau: Ah! Thanks.

PR #376, opened 2025-09-26
Fix: fix short buffer errors

ppotapov-aws:


interop/klr/gather.c
```diff
@@ -1676,8 +1676,9 @@ PyObject* specialize(struct kernel *k, PyObject *args, PyObject *kws, PyObject *
     if (first) first = false;
     else append(&json, ",");
 
-    snprintf(buf, sizeof buf, "\"%s:%d:%d %s\"", m->file, m->line, m->col, m->message);
+    asprintf(&buf, "\"%s:%d:%d %s\"", m->file, m->line, m->col, m->message);
```
govereau, line 1679: If there is an error, `buf` can be NULL.
Should we check this?

govereau (approved):

PR #382, opened 2025-09-29
chore: Update Lean to 4.23

seanmcl:


interop/klr/klir_ast.hpp
```diff
@@ -10,10 +10,21 @@ Written by the KLR Contributors (https://github.com/leanprover/KLR)
 
 #include "klir_common.hpp"
```
seanmcl, line 12: no idea what happend to these files

PR #384, opened 2025-09-30
Fix: remove classes that don't inherit from NKI or Enum

ppotapov-aws:


interop/klr/gather.c
```diff
@@ -1459,6 +1459,11 @@ static void class(struct state *st, lean_object *name, struct _stmt *s) {
       error(st, "NKI classes must inherit from either Enum or NKIObject");
       return;
     }
+  } else {
```
govereau, line 1462: Did you check this on a kernel?
I was worried that we would have tons of errors?

govereau (approved):

PR #391, opened 2025-10-01
fix: reduce cmd was not properly parsing

ppotapov-aws:
Since we are parsing objects now, reduce_cmd comes through as an object and not string. Let's fix that

### DIFF
--- a/KLR/Trace/FromNKI.lean
+++ b/KLR/Trace/FromNKI.lean
@@ -323,8 +323,16 @@ instance : FromNKI ActivationFunc where
 
 instance : FromNKI AccumCmd where
   fromNKI? t :=
-    let err := .error "expecting activation function type"
+    let err := .error s!"expecting accumulator command (idle, reset, reduce, reset_reduce), got {repr t}"
     match t with
+    | .object `nki.isa.reduce_cmd vs
+    | .object `neuronxcc.nki.isa.reduce_cmd vs =>
+      match AA.lookup? vs "name" with
+      | some (.string "idle") => return .Idle
+      | some (.string "reset") => return .Zero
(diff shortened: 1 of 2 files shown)


KLR/Trace/FromNKI.lean
```diff
@@ -323,8 +323,15 @@ instance : FromNKI ActivationFunc where
 
 instance : FromNKI AccumCmd where
   fromNKI? t :=
-    let err := .error "expecting activation function type"
+    let err := .error s!"expecting accumulator command (idle, reset, reduce, reset_reduce), got {repr t}"
     match t with
+    | .object `neuronxcc.nki.isa.reduce_cmd vs =>
+      match vs.find? (·.1 == "name") with
```
govereau, line 329: `vs.lookup` may be better here.

KLR/Trace/NKI.lean
```diff
@@ -50,6 +50,7 @@ def NKIEnv : List (Name × Term) :=
   , const_var (nisa "tensor_engine")
   , const_var (nisa "vector_engine")
   , const_var (nisa "scalar_engine")
+  , const_var (nisa "reduce_cmd")
```
govereau, line 53: Hmm, why do we need this here?
I would think the environment would already get a definition from the class.
Also, doesn't this break the constructor? e.g.
```
x = reduce_cmd('x', 1)
```

govereau (changes requested):

KLR/Trace/NKI.lean
govereau: Hmm, something went wrong. We should have a definition for `reduce_cmd` caused by gathering the `reduce_cmd.idle` value.

PR #394, opened 2025-10-01
Fix: activation bias type should be just tensor

ppotapov-aws:


interop/klr/klir_common.cpp
```diff
@@ -68,7 +68,7 @@ String String_des(FILE *in) {
     throw std::runtime_error("expecting String");
 
   String result = res;
-  free(res);  // std::string make a copy
+  free(res); // std::string make a copy
```
govereau, line 71: Why is this line deleted?

govereau: For context, Hayden asked for this line saying the `std::string` makes a copy. I assumed he was correct.

govereau (approved):

PR #401, opened 2025-10-06
fix: naming discrepancy

ppotapov-aws:
``` sh
(ENV) ➜  NeuronKernelInterface git:(mainline) ✗ strings attention.klir | grep k_out
gk_out.3
gk_out.3
gk_out.3
gk_out.3
```
Prior run:
```
=========================================================================================================================================================================== slowest 5 durations ============================================================================================================================================================================
0.13s call     test/frontend/test_attention_token_gen_cascaded.py::test_kernel[4-1-5-1024-10240-128-True-True-True-True-False-False-False-bfloat16-None]

(2 durations < 0.005s hidden.  Use -vv to show these durations.)
============================================================================================================================================================================ 1 passed in 1.82s =============================================================================================================================================================================
(ENV) ➜  NeuronKernelInterface git:(mainline) ✗ strings attention.klir | grep k_out
hk_out.18
hk_out.18
ik_out.470
ik_out.470
```


govereau:
OK, interesting. This makes senses. The compiler is generating fresh names and these should not prevent us from reusing those names in later compiler passes. This happens because we have everything in one computation here:

  https://github.com/leanprover/KLR/blob/main/KLR/Compile.lean#L44-L49

I think a better way to do this would be to either split the `PassM` computations up, or maybe create some kind of `reset` function in `PassM` to clear the state.

waahm7 (approved):

